PR #18, opened 2023-11-28
Tests and Efficiency

mmschlk:
# Tests
This PR adds tests to all approximators using the `DummyGame`. 
- tests for `InteractionValues`
- tests for `PermutationSamplingSTI`
- tests for `RegressionFSI`
- tests for `ShapIQ`

# Refactoring of InteractionValues
 Changed the internal representation of interaction scores to be a vector of length `n_interactions` instead of mostly empty arrays. This makes it easier to approximate interaction scores for higher player counts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@
 # SHAP-IQ: SHAP Interaction Quantification
 > An interaction may speak more than a thousand main effects.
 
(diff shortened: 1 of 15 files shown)


mmschlk:
Bug in getitem still persists interaction tuple needs to be sorted and test case

PR #22, opened 2023-12-01
Add nSII and SII Regression Estimator

mmschlk:
# nSII Index
Adds functionality to compute nSII values from SII estimators or to transform SII `InteractionValues` to nSII ones.

# Regression SII
Adds regression estimator for estimating SII values.


mmschlk:
LGTM!

mmschlk:
LGTM too!

PR #114, opened 2024-04-12
remove warning for adding two indices (closes 88)

mmschlk:
Removed the warning instead of suppressing, as this actually is very reasonable and default behavior of adding two `InteractionValues` objects. Especially when the object becomes more sparse.


hbaniecki (approved):
lgtm

PR #123, opened 2024-04-16
96 add permutation sampling for sv

kolpaczki:
Added Permutation Sampling (ApproShapley) by Castro et al. (2009) together with tests:
- Each coalition of a permutation is only evaluated once to increase efficiency.
- The empty and full coalition are evaluated in the beginning and the values are stored to save further budget usage.


mmschlk (approved):
LGTM

PR #135, opened 2024-04-19
71 add owen sampling sv approx

kolpaczki:
Added Owen Sampling (Okhrati & Lipani, 2020) for the Shapley value including:

- Hyperparameter M to set the number of anchor points at which the integeral of each player is palpated between 0 and 1 in equidistant steps.
- Helper method to generate these points (tested).
- Tests including multiple instanciations of M.
- The budget is split evenly acroos all players and anchor points: each player and each anchor poitn is approximated with the same number of samples.


mmschlk (approved):
LGTM ... the for loops look scary but do only create an array so should be fast enought

PR #150, opened 2024-05-02
Refactor and clean benchmark games

hbaniecki:
**Things to consider/discuss**:
1. Benchmark/games rely heavily on `scikit-learn`. I can only propose moving scikit-learn to package dependencies and then moving its imports (e.g. accuracy, random forest) outside the function calls.
2. In some places `random_state=None`, in other `random_sate=42`, and in some there was no `random_state` at all. I just tried to set `random_state=42` everywhere. An alternative probably is to set `random_state=None`.
3. In spirit of unification, I changed `empty_prediction` to `normalization_value` across different benchmark/games. (annoying? can be reverted) 
    - Also, one test does not pass, but shouldn't `normalize=False` make `game_values[0] == 0.0` instead of `0.1`? image
    - In `games/base.py` there is image

**Other minor**:
- Should ImageClassifier and SentimentAnalysis inherit from LocalExplanation and not Game?
- Default `cluster_method` differs between AdultCensus and BikeSharing / CaliforniaHousing.
- We should rethink class names like CaliforniaHousingTreeSHAPIQXAI.


mmschlk (approved):
reverted the normalization_value unification but made the variables more expressive (intentionally different to be in context of the benchmark games).

PR #167, opened 2024-06-05
Sv bugfix in tree

mmschlk:
Adds "SV" index to TreeSHAP-IQ such that one no longer has to do "SII" and `max_order = 1`.


hbaniecki (approved):
lgtm

PR #185, opened 2024-08-13
182 add core of order 1 to exactcomputer

Advueu963:
This PR adds the **egalitarian-least-core** as a tool to the exact computer only relying on the scipy package.


shapiq/exact.py
```diff
@@ -830,6 +834,32 @@ def probabilistic_value(self, index: str, *args, **kwargs) -> InteractionValues:
...
+
+            # Compute egalitarian least-core
+            egalitarian_vector, subsidy = core.egalitarian_least_core(
+                n_players=self.n,
+                game_values=self.game_values,
+                coalition_lookup=self.coalition_lookup,
+                grand_coalition_tuple=self._grand_coalition_tuple,
+            )
+        else:
+            raise ValueError(f"Index {index} not supported")
```
mmschlk, line 855: I don't like this if/else structure here at the moment. Let me elaborate:

1. First, I don't like the check with "ELC" as only ELC is acceptable here, any other index will lead to an error. The user should not have to specify any `index `here. So we could just drop the `index `parameter entirely and leave args and kwargs as they are there now.
2. Second, I am not 100% sure if we cannot do something if `if self.baseline_value != 0:`. Can we re-normalize here? If not then we need to error out, yes. But if yes, I would like to do this here.
3. Third, the "Please set game_fun.normalize=True!" is a bit misleading. If a `game_fun` is an instance of `Game` then this will not resolve the problem. Setting the attribute will not trigger a re-normalization of the game. The normalize parameter must be set to `True` at the initialization of the game. So we would at least update the error text here to reflect that. :)

tests/test_exact_computer.py
```diff
@@ -79,6 +79,52 @@ def test_exact_computer_on_soum():
...
+        egalitarian_least_core = exact_computer.compute_egalitarian_least_core(index)
+        # Assert efficiency for Egalitarian Least Core
+        assert (np.sum(egalitarian_least_core.values) - predicted_value) ** 2 < 10e-7
+
+
+@pytest.mark.parametrize(
+    "index, order",
+    [("ELC", 1)],
+)
+def test_exact_elc_computer_call(index, order):
```
mmschlk, line 114: What should the ELC min_order be in the `InteractionValues` object after creation. I would like to add a check to that. I think it should be min_order == 1, right?

tests/test_exact_computer.py
```diff
@@ -79,6 +79,52 @@ def test_exact_computer_on_soum():
...
+    soum = SOUM(n, n_basis_games=10, normalize=True)
+    exact_computer = ExactComputer(n_players=n, game_fun=soum)
+    interaction_values = exact_computer(index=index, order=order)
+    if order is None:
+        order = n
+    assert interaction_values is not None  # should return something
+    assert interaction_values.max_order == order  # order should be the same
+    assert interaction_values.index == index  # index should be the same
+    assert interaction_values.estimated is False  # nothing should be estimated
+    assert interaction_values.values is not None  # values should be computed
```
mmschlk, line 126: I would also like a check to baseline_value == 0

tests/test_exact_computer.py
```diff
@@ -79,6 +79,52 @@ def test_exact_computer_on_soum():
...
+        exact_computer = ExactComputer(n_players=n, game_fun=soum_normalized)
+
+        predicted_value = soum_normalized(np.ones(n))[0]
+
+        egalitarian_least_core = exact_computer.compute_egalitarian_least_core(index)
+        # Assert efficiency for Egalitarian Least Core
+        assert (np.sum(egalitarian_least_core.values) - predicted_value) ** 2 < 10e-7
+
+
+def test_exact_elc_on_normalized_soum():
```
mmschlk, line 94: General comment to the tests: I would like to have semantic checks of the core in a separate test file (not in the `test_exact_computer.py` file) and in the `test_exact_computer.py` an additional check that sees that it works with the exact one.

mmschlk (changes requested):
Awesome stuff!!!! I added some small changes.

mmschlk (approved):

PR #242, opened 2024-10-04
183 make games be also callable with a tuplelist of tuples

Advueu963:
Added functionality for Game to be called with both list  of tuples or just a tuple.
Depending on optional player names given to the game, the elements of the tuple(s) can be strings.


shapiq/games/base.py
```diff
@@ -157,7 +163,106 @@ def is_normalized(self) -> bool:
...
+                        f"It should have a length of {self.n_players}"
+                    )
+                coalitions = coalitions.reshape((1, self.n_players))
+            if coalitions.shape[1] != self.n_players:
+                raise TypeError(
+                    f"The number of players in the coalitions ({coalitions.shape[1]}) does not match "
+                    f"the number of players in the game ({self.n_players})."
+                )
+            return coalitions
+        elif isinstance(coalitions, str):
```
mmschlk, line 244: This logic could also work with an integer (i.e. just the id of a player) dunno if we want this actually. Could be maybe misleading if users mean "get me the first coalition". Maybe this would be better served with a _getitem_ or something.

mmschlk (changes requested):
I marked a couple of nitpicks. 🤓

mmschlk (approved):
Very nice! :) I like this much better just raising the error. :)

shapiq/games/base.py
mmschlk, line 1: I love the examples!

PR #280, opened 2024-11-22
Create poster.pdf

hbaniecki:


mmschlk (commented):
poster is not finished, yet

mmschlk:
merging this but we need to change the poster once its finsihed

Issue #275, opened 2024-11-13
Remove shap as optional dependency


mmschlk:
Currently shap is used as a dependency for creating some of the plots in `shapiq`. Keeping shap as an optional dependency makes updating `shapiq` to newer Python versions and allowing shapiq to work with newer numpy versions quite stressful and tricky. 

I think removing shap as imports anywhere in the project could be a good idea. For this we would need to re-create the plots in our own version. This could be more or less effort.

hbaniecki:
Good idea.

PR #296, opened 2025-01-05
275 remove shap as optional dependency

Advueu963:
This PR attempts to remove shap as an optional dependency and closes issue #275 .

Currently, _bar_,_waterfall_ and _force_ plots depend on their corresponding shap implementation.
This has been removed, by re-using the necessary code but eliminating any conversion functions currently needed.
In particular, now the plots only depend on the `InteractionValues` object.


mmschlk (approved):
I think now this PR can be merged. Some minor things might still be subject to change but this is a very nice contribution!

Issue #286, opened 2024-12-04
Can't instanciate TreeExplainer when a tree in LightGBM Classifier model has n_features_in_tree = 1


CharlesCousyn:
I have a LightGBM classifier working well.
But when i execute:
```
explainer  = shapiq.TreeExplainer(
    model=model.named_steps["classifier"], # My LightGBMClassifier
    max_order= 1,
    min_order= 1,
    index="k-SII",
    class_index=0
)
```
It fails with:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[6], line 1
----> 1 explainer  = shapiq.TreeExplainer(
      2     model=model.named_steps[\"classifier\"],
      3     max_order= 1,
      4     min_order= 1,
      5     index=\"k-SII\",
      6     class_index=0
      7 )
      8 shap_values, _ = explainer.explain(transformed_dataset)
     10 # Compute shap values
     11 # shap_values = feature_metrics.get_shap_values(model.named_steps[\"classifier\"], x, force_use_all_data=True)
     12 
     13 # Convert shap into readable dataframe

File ~/Code/isa-ml/isa-ml-jobs/.venv/lib/python3.10/site-packages/shapiq/explainer/tree/explainer.py:73, in TreeExplainer.__init__(self, model, max_order, min_order, index, class_index, **kwargs)
     70 self._class_label: Optional[int] = class_index
     72 # setup explainers for all trees
---> 73 self._treeshapiq_explainers: list[TreeSHAPIQ] = [
     74     TreeSHAPIQ(model=_tree, max_order=self._max_order, index=index) for _tree in self._trees
     75 ]
     76 self.baseline_value = self._compute_baseline_value()

File ~/Code/isa-ml/isa-ml-jobs/.venv/lib/python3.10/site-packages/shapiq/explainer/tree/explainer.py:74, in <listcomp>(.0)
     70 self._class_label: Optional[int] = class_index
     72 # setup explainers for all trees
     73 self._treeshapiq_explainers: list[TreeSHAPIQ] = [
---> 74     TreeSHAPIQ(model=_tree, max_order=self._max_order, index=index) for _tree in self._trees
     75 ]
     76 self.baseline_value = self._compute_baseline_value()

[...]

PR #310, opened 2025-01-30
Fix TreeExplainer instanciation when a tree in LightGBM Classifier model has n_features_in_tree = 1

CharlesCousyn:
Hello,

I am opening this PR to fix the bug mentioned [here](https://github.com/mmschlk/shapiq/issues/286) for the TreeExplainer.

Let me know id it does fix the problem correctly!

### DIFF
--- a/shapiq/explainer/tree/treeshapiq.py
+++ b/shapiq/explainer/tree/treeshapiq.py
@@ -409,7 +409,7 @@ class TreeSHAPIQ:
             self.subset_ancestors_store[order] = subset_ancestors
 
             # If the tree has only one feature, we assign a default value of 0
-            if self._n_features_in_tree == 1:
+            if self.n_interpolation_size == 1:
                 self.D_store[order] = np.array([0])
             else:
                 self.D_store[order] = np.polynomial.chebyshev.chebpts2(self.n_interpolation_size)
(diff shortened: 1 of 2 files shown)


mmschlk:
Hey @CharlesCousyn, thank you so much for taking care of this! This is very appreciated!

One quick thing. Would you be able to add a very small unit test as a minimal example that would error in the current version without your fix but succeeds with your fix? This way the fix is future proof? 

I think it would be good if the TreeExplainer is used to explain a very small "model" with only one feature. For this you could use a Custom Tree (here is some info on the custom tree: https://shapiq.readthedocs.io/en/latest/notebooks/tree_notebooks/treeshapiq_custom_tree.html).

Best Max!

mmschlk (changes requested):
Thank you so much!

Could you add a very small unit test as mentioned in the comment?

CharlesCousyn:
@mmschlk Done!

mmschlk:
> @mmschlk Done!

Amazing! I will check it out and merge it! :) Thank you again! :)

mmschlk (approved):
Tested, it on the current main and the bug is fixed. :)

mmschlk:
This PR fixes and closes #286

Issue #316, opened 2025-02-12
Fix bar and force plotting


Advueu963:
As mentioned in issue #315 there is an issue with the aggregation of multiple interaction value objects in the `bar_plot` function.
This is due to a wrong ordering in the values object given to the plot function.
Additionally, there is an issue with the `force_plot` also having a wrong ordering of the values given to the function.

Issue #315, opened 2025-02-12
Global Summary Plot: Mis-ordered importances and aggregation bug


danielarifmurphy:
Hi shapiq team - great job on the package!

I was looking to reproduce the [global interaction importance summary plot](https://shapiq.readthedocs.io/en/latest/notebooks/tree_notebooks/treeshapiq_lightgbm.html#Global-Feature-(Interaction)-Importance) found in the TreeSHAP-IQ for LightGBM tutorial.

Running the same code (with version `1.2.0`) generated the following plot, which differs from that found in the docs:

![Image](https://github.com/user-attachments/assets/c30a3cd3-4e53-41cd-9313-5ff0710e8e42)

A couple of points:

1. The new bar plot logic [aggregates interaction values](https://github.com/mmschlk/shapiq/blob/de2d37cf34e7097c23f7985484af4692f1e36b2b/shapiq/plot/bar.py#L223) across samples by taking the [simple mean](https://github.com/mmschlk/shapiq/blob/de2d37cf34e7097c23f7985484af4692f1e36b2b/shapiq/interaction_values.py#L883-L884) rather than the mean of _absolute values_ as was the case previously when using the [`shap` bar plot.](https://github.com/mmschlk/shapiq/commit/34a86e3970447f694b7418b682aab085332b1ac6?diff=split&w=0#diff-effb76329b97ef7ad01c326c334f403e37e69491d11bacf49cfb8dacb3fb19ecL62:~:text=ax%20%3D%20shap.plots.bar(explanation%2C%20**kwargs%2C%20show%3DFalse)). Perhaps this was intentional but having an option to specify global absolute effect on predictions would be useful, as we get some averaging out effects with the simple mean.
2. Regardless of 1. it seems the plot output has incorrect ordering of importance.
3. The `min_order` parameter currently has no effect - it gets [instantiated in the explainer](https://github.com/mmschlk/shapiq/blob/de2d37cf34e7097c23f7985484af4692f1e36b2b/shapiq/explainer/tree/explainer.py#L69) but is not used thereafter. Might be worth at least adding an option to filter the summary plot such that it displays e.g. only order 2 interactions.

Thanks again for all your great work!

Thanks,
Dan

PR #317, opened 2025-02-12
316 fix bar and force plotting

Advueu963:
This PR fixes the bugs described in issues #315 and #316 by fixing the ordering of the values.

Also, the LightGBM notebook was updated with the new interface to the functions.


mmschlk (approved):
Yes, very nice!

Issue #291, opened 2024-12-14
Python 3.13


mmschlk:
We need to update to python 3.13 and could drop 3.09. 

The current hold-up is that our unit tests for plotting depend on shap which appears not to install with the older numpy or scipy versions.

PR #318, opened 2025-02-14
291 Change the Supported Python Versions to 3.10-3.13

Advueu963:
TLDR: This PR tackles issue #291 to change the range from supported versions from 3.9-3.11 to 3.10-3.13.

In more detail the following steps were conducted

1. Any mentions of 3.9 were changed to 3.10 in documentations and GitHub-actions.
2. The type annotations `Union[X,Y]` were changed to `X | Y`.
3. The type annotations `Optional[X]` were changed to `X | None`.
4. `pd.option_context("future.no_silent_downcasting", True): ...` in tree conversion functions were removed, as it seemed to not be necessary anymore


mmschlk:
I also added support for Python 3.13. Unfortunately, not all libraries already work with 3.13. So I had to update som tests and conditional imports in the requirements file (used for the test matrix) and skip some tests.

Currently these packages do not support 3.13:
- tabpfn
- tensor flow
- tf-keras

mmschlk (approved):

Issue #322, opened 2025-02-14
Bump version to development, update the changelog


hbaniecki:
The main branch says `1.2.0` while it denoted the most recent version on PyPI https://pypi.org/project/shapiq/#history

https://github.com/mmschlk/shapiq/blob/4280a7d36350adc3aff4685d5659dabf31156217/shapiq/__init__.py#L5

Change this e.g. to `1.2.0.9000`

Also, update the changelog https://github.com/mmschlk/shapiq/blob/main/CHANGELOG.md

Issue #319, opened 2025-02-14
shapiq.Explainer breaks when index = "SV" with approximator != "auto"


hbaniecki:
This line breaks for `shapiq.Explainer(model, data, index="SV", approximator="permutation")` because SV approximators have no `max_order`.

https://github.com/mmschlk/shapiq/blob/4280a7d36350adc3aff4685d5659dabf31156217/shapiq/explainer/tabular.py#L270

Potential solution: Add `**kwargs` to SV approximators.

mmschlk:
Oh, I see this might happen also for SVARM? I will build a patch release on Monday for the plot bugs fixed today and fix this here as well.

hbaniecki:
this will happen for all SV approximators in the Explainer interface

mmschlk:
Uff...

PR #323, opened 2025-02-14
partial fi(x)es for #324, #322, #319

hbaniecki:
partial fi(x)es for #324, #322, #319


mmschlk (approved):
Added SV support to svarm and unbiased kernelshap and added a test which fails without the fix

PR #327, opened 2025-02-20
Fix Package Import without Write Access

mmschlk:
This PR simply wraps the creation of the data folder in a function which will be called upon loading the datasets and not at init time of the package. This means that if the package is installed in environments without write access also possible to import.


pwhofman (approved):
Works on my machine!

PR #335, opened 2025-03-16
bug fix for stacked_bar plot

Advueu963:
**TLDR**: The fix comprises updating the `get_n_order_values` in `Interactionvalues` as it necessary for the stacked bar plot to work properly.

Currently the stacked_bar plot inflates the bars incorrectly and thus makes the interpretation difficult.
If given the following simple game setup
```
import matplotlib.pyplot as plt
import numpy as np

from shapiq import ExactComputer, powerset, Game
from shapiq.games.benchmark import SOUM

class SimpleGame(Game):
    def __init__(self) -> None:
        super().__init__(n_players=3, normalize=True, normalization_value=0)

    def value_function(self, coalitions: np.ndarray) -> np.ndarray:
        coalition_values = {
            (): 0,
            (0,): 0,
            (1,): 0,
            (2,): 0,
            (0, 1): 100,
            (0, 2): 100,
            (1, 2): 100,
            (0, 1, 2): 300,
        }

        values = np.array([coalition_values[tuple(np.where(x)[0])] for x in coalitions])

        return values

game = SimpleGame()


exact_computer = ExactComputer(n_players=game.n_players, game=game)

moebius = exact_computer("Moebius")
moebius.plot_stacked_bar(show=True)
```
For this game the interaction values of (0,1), (0,2) and (1,2) equal 100 and the rest is zero.
While this is correctly computed the problem lies in the visualisation:
image

With the fix it correctly becomes:
image


mmschlk (changes requested):
Could you document this in the CHANGELOG.md as well?

mmschlk (approved):
<3

Issue #355, opened 2025-03-26
Disallow `budget=None` in explainer and/or other occurrences


mmschlk:
Many functions in shapiq rely on a `budget: int | None = None` parameter, which limits the number of model calls for approximating interactions. If users provide `budget=None` (the default) the budget will be set to 2^n_features/n_players. This is happening with a warning to let users know that many model calls are going to happen:

```python
  if budget is None:
      budget = 2**self._n_features
      if budget > 2048:
          warnings.warn(
              f"Using the budget of 2**n_features={budget}, which might take long\
                        to compute. Set the `budget` parameter to suppress this warning."
          )
```

As it turns out, this is not a very nice way of dealing with such an integral part of our API. All approximatiors currently rely on the budget. And the estimation quality can be drastically different from use-case to use-case. This means setting a default value never is good fit for all (or actually many) use cases. By defaulting to None users do not expect this parameter to be as important as it actually is.

**Proposed solution**: Make `budget: int` instead of `budget: int | None = None` and remove possibility to set budget to `None`.

mmschlk:
This happened at https://github.com/mmschlk/shapiq/issues/330 and at https://github.com/mmschlk/shapiq/issues/354.

PR #356, opened 2025-03-27
355 disallow budgetnone in explainer andor other occurrences

Advueu963:
This PR deals with issue #355, namely that the `budget` parameter was not highlighted relevant, through the decision of `budget = None`. Yet it is a crucial part of Explanations, with coalition samples.

Thus this PR changes the parameter to be necessary.

In this context tests needed to be updated to account for the API change.

While these changes do make the usage somewhat more "complicated,” they highlight the importance of coalition samples and their implications for memory and time consumption.

### DIFF
--- a/shapiq/explainer/tabular.py
+++ b/shapiq/explainer/tabular.py
@@ -188,14 +188,11 @@ class TabularExplainer(Explainer):
         Args:
             x: The data point to explain as a 2-dimensional array with shape
                 (1, n_features).
-            budget: The budget to use for the approximation.
+            budget: The budget to use for the approximation. It indicates how many coalitions are sampled, thus high values indicate more accurate approximations, but induce higher computational costs.
+                In particular one should be aware of the exponential growth of the number of coalitions with the number of features i.e. 2**(n_players).
+                We abstain from giving a general recommendation for the budget, as it depends on the specific use case and the desired accuracy.
             random_state: The random state to re-initialize Imputer and Approximator with. Defaults to ``None``.
         """
-        if budget > 2048:
-            warnings.warn(
-                f"Using the budget of 2**n_features={budget}, which might take long\
-                          to compute. Set the `budget` parameter to suppress this warning."
-            )
         if random_state is not None:
             self._imputer._rng = np.random.default_rng(random_state)
             self._approximator._rng = np.random.default_rng(random_state)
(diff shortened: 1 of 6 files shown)


shapiq/explainer/tabular.py
```diff
@@ -181,24 +181,21 @@ def __init__(
...
-            if budget > 2048:
-                warnings.warn(
-                    f"Using the budget of 2**n_features={budget}, which might take long\
-                              to compute. Set the `budget` parameter to suppress this warning."
-                )
+        if budget > 2048:
+            warnings.warn(
+                f"Using the budget of 2**n_features={budget}, which might take long\
+                          to compute. Set the `budget` parameter to suppress this warning."
+            )
```
mmschlk, line 198: The warning currently reads 2**n still. I would remove the warning, since now the users set the budget themselves. This keeps it clean. :)

tests/tests_explainer/test_explainer_tabular.py
```diff
@@ -111,11 +112,11 @@ def test_init_params_approx_params(dt_reg_model, background_reg_data, approximat
     explainer = TabularExplainer(
         approximator=approximator, model=dt_reg_model, data=background_reg_data, max_order=max_order
     )
-    iv = explainer.explain(background_reg_data[0])
+    iv = explainer.explain(background_reg_data[0],budget=BUDGET_NR_FEATURES)
     assert iv.__class__.__name__ == "InteractionValues"
 
 
-BUDGETS = [2**5, 2**8, None]
+BUDGETS = [2**3, 2**5, BUDGET_NR_FEATURES]
```
mmschlk, line 119: Atm, I am unsure if the test specifically was intended to also run with budget higher than 2**n if it was the case then I would also do this again here. :)

mmschlk (changes requested):

mmschlk (changes requested):
Also, this change should be properly documented in the docstrings and in the changelog.

shapiq/explainer/tabular.py
Advueu963: Removed.

Advueu963:
The issues should be resolved, but I am not sure regarding „updating docstrings“. I now added more information on the parameter, but maybe you had an different idea @mmschlk .

mmschlk (approved):

Issue #357, opened 2025-03-27
Add API Example Notebook for Upset plots


mmschlk:
We recently introduced the upset plot to shapiq to visualize interactions (after Hendrik Strobelt told us this is the way). However, we did not yet add information on how to use and interpret the values. For this an API notebook would be pretty good.

mmschlk:
Done in #366

PR #366, opened 2025-04-01
Enforce "PTH" Linting Rule and Adds UpSet API Example Notebook

mmschlk:
This PR enforces the `"PTH"` rule and adds an API Example notebook showing how to use UpSet plots for explanations. This also marks the first example of how API example notebooks could look like (though maybe the sklearn imports/runs could be replaced with loading an InteractionValues object from scratch).


mmschlk:
This PR closes #357.

PR #379, opened 2025-05-07
Add the SPEX approximator to shapiq

justinkang221:
This pull request adds the SPEX method for computing attributions over large input spaces with up to $n \approx 1000$ features. SPEX was developed at UC Berkeley, and uses ideas from information and coding theory to ensure fast and efficient computation of interactions. 

The article detailing the SPEX algorithm is available here:
https://arxiv.org/pdf/2502.13870
A version of this article will appear at ICML 2025. 

image


mmschlk:
@justinkang221 amazing what you guys did here! ❤️

I will go through the PR some time this weekend probably! ☺️

justinkang221:
Sounds good @mmschlk. One thing I wanted to highlight is that I added  "sparse-transform" and "galois" to the dependencies. I know you mentioned you wanted to mark them as optional, so feel free to make that change, I wasn't sure how best to handle it.

mmschlk (approved):
This is so super nice! I don't see any reason why we should not merge directly! The only thing, I would add is a tutorial/introduction notebook which goes a bit beyond the current additions to the existing notebooks. But this I think we can tackle also at a later date. :)

mmschlk:
> Sounds good @mmschlk. One thing I wanted to highlight is that I added  "sparse-transform" and "galois" to the dependencies. I know you mentioned you wanted to mark them as optional, so feel free to make that change, I wasn't sure how best to handle it. 

It seems like the dependencies are pretty lightweight and seem to install already in all python versions in the tests. So I would leave it as is for now and monitor how it goes. :)

mmschlk:
I will do minor things like adding it to changelog and adding your method more prominently into the docs next.

PR #397, opened 2025-06-04
Add `Game.compute` function and related test to the Game class

IsaH57:
This pull request adds a `Game.compute()` function to the `shapiq.Game` class. After computing the game values, it returns the normalized game values, the coalition_lookup, and the normalization_value. Notably, using the compute function does not alter the state of the `Game` object. This PR also adds an utility function `generate_interaction_lookup_from_coalitions` for creating an interaction lookup dict from an arbitrary coalition array.


shapiq/games/base.py
```diff
@@ -419,6 +419,30 @@ def precompute(self, coalitions: np.ndarray | None = None) -> None:
         self.coalition_lookup = coalitions_dict
         self.precompute_flag = True
 
+    def compute(
+        self, coalitions: np.ndarray | None = None, *, return_normalization: bool = False
```
mmschlk, line 423: could you actually always return the normalization_value as the third thing and not normalize the values in line 440. Then this function would be more usable because it's the "raw" computation.

mmschlk (changes requested):

shapiq/games/base.py
```diff
@@ -419,6 +419,26 @@ def precompute(self, coalitions: np.ndarray | None = None) -> None:
...
+    def compute(
+        self, coalitions: np.ndarray | None = None
+    ) -> tuple[np.ndarray, dict[tuple[int, ...], int], float]:
+        """Compute the game values for all or a given set of coalitions.
+
+        Args:
+            coalitions: The coalitions to evaluate.
+
+        Returns:
+            A tuple containing:
```
mmschlk, line 431: I deleted the return types because sphinx will display automatically and then we do not have to maintain two parts of the code. :)

mmschlk (approved):

Issue #399, opened 2025-06-15
Add beeswarm plot to SHAP-IQ


annprzy:
Beeswarm plots are useful in visualizing dependencies between feature values and feature importance. A beeswarm plot could be adapted from the SHAP library by sub-dividing the y-axis for each interaction term:
- An interaction term (e.g., "Feature A x Feature B") will occupy a "block" on the y-axis.
- This block will be subdivided into multiple rows: one for each feature within the interaction.
- For the interaction (A, B):
  * There will be a sub-row for Feature A. The x-position of every dot in this row will be the interaction value for that sample, but the color of the dot will be determined by the value of Feature A.
  * There will be another sub-row for Feature B. The x-position of the dots is the same, but the color will be determined by the value of Feature B

Example of such plot could look like this:

![Image](https://github.com/user-attachments/assets/2b92d7f0-8e2a-4035-a847-f4b62fa549c8)

annprzy:
@mmschlk Thank you! Regarding PR, I think everything is almost ready and I will open it later today. Just maybe a quick question - I've noticed you uploaded some tutorials for the visualizations (upset and si graph plots). Would you like me to add one for the beeswarm plot as well or these are just for less-known/newer plot types that include more customization options?

mmschlk:
@annprzy, this would be amazing, if you have the time! :)

mmschlk:
Closed in #406.

PR #406, opened 2025-06-17
Add beeswarm plot to SHAP-IQ

annprzy:
This PR addresses #399, adapting SHAP beeswarm plot to handle interactions and adding it to the `plot` module.

New features:
- `shapiq/plot/beeswarm.py`: adapts SHAP beeswarm plot to handle interactions
- `tests/tests_plots/test_beeswarm.py`: adds test for beeswarm plot

Updated:
- `shapiq/__init__.py`: adds beeswarm plot to the list of imports and `__all__` attribute
- `shapiq/plot/__init__.py`: adds beeswarm plot to the list of imports and `__all__` attribute


mmschlk:
@annprzy, would it technically be possible to work without the skimage import? I see that you only require the conversion of color spaces there. I would love to not blow up the dependencies of shapiq too much and we already use [`colour`](https://pypi.org/project/colour/) or `matplotlib` which probably can also do what you need to do with skimage. Maybe the code would also work without the conversions even?

Second, I know you asked this in the issue, but if you do have some additional time for this, would you be able to add a quick tutorial for the plot under `api_examples` like the [upset plot](https://github.com/mmschlk/shapiq/blob/main/examples/api_examples/visualization_upset_plot.ipynb)? This really does not need to be too elaborate but some quick showcase of the plot? 

Thank you for this PR! :)

annprzy:
@mmschlk I simplified the colormap generation (skimage is not longer needed) and added the tutorial. Let me know your thoughts on this and whether any changes are needed.

PR #424, opened 2025-08-05
Remove zeros from interaction lookup, check in test

### LINKED ISSUE #369: Moebius_converter adds many value 0 interactions during conversion
During the conversion from a sparse Möbius representation to another index, sometimes many value 0 interactions are added to the returned InteractionValues.

This is because interactions are added to the interaction_dict, even if their coefficient is 0. 

- https://github.com/mmschlk/shapiq/blob/f3e9e7f6e7d4a8b0c56ba33f9adf22f8932a607a/shapiq/game_theory/moebius_converter.py#L125-L130
- https://github.com/mmschlk/shapiq/blob/f3e9e7f6e7d4a8b0c56ba33f9adf22f8932a607a/shapiq/game_theory/moebius_converter.py#L199-L204
- https://github.com/mmschlk/shapiq/blob/f3e9e7f6e7d4a8b0c56ba33f9adf22f8932a607a/shapiq/game_theory/moebius_converter.py#L275-L280

Alternatively, the returned InteractionValue could be sparsified upon return.

CloseChoice:
## Motivation and Context
---
closes #369.
Prevented zeros from being written to the `interaction_lookup`.
I added this to the test, but it seems like only the change in [here](https://github.com/mmschlk/shapiq/compare/main...CloseChoice:shapiq:369-remove-0s-from-interaction-lookup?expand=1#diff-fb4550248bea0697dcb027a2fec245f5fdf2b3eb42edc89b34f64b30c0e34841R166-R169) runs through the tests. If more thorough tests are desired, let me know.

## Public API Changes

-   [x] No Public API changes

---

## How Has This Been Tested?
see the changes in the tests.
---

## Checklist

---

### DIFF
--- a/tests/shapiq/tests_unit/tests_game_theory/test_moebius_converter.py
+++ b/tests/shapiq/tests_unit/tests_game_theory/test_moebius_converter.py
@@ -27,7 +27,7 @@ def test_soum_moebius_conversion():
             # Assert efficiency
             assert (np.sum(shapley_interactions.values) - predicted_value) ** 2 < 10e-7
             assert (shapley_interactions[()] - emptyset_prediction) ** 2 < 10e-7
-            for _k, v in moebius_converter._computed.items():
+            for v in moebius_converter._computed.values():
                 # check that no 0's are in the interaction lookup (except for the empty set, which is the first entry)
                 interaction_lookup = v.interaction_lookup
                 assert all(v != 0 for idx, v in enumerate(interaction_lookup.values()) if idx > 0)
--- a/tests/shapiq/tests_unit/tests_game_theory/test_moebius_converter.py
+++ b/tests/shapiq/tests_unit/tests_game_theory/test_moebius_converter.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import numpy as np
+import pytest
 
 from shapiq.game_theory.moebius_converter import MoebiusConverter
 from shapiq_games.synthetic.soum import SOUM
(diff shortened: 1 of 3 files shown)


mmschlk:
Hi @CloseChoice, welcome to shapiq! Happy to have you here! :)
Thank you for the PR! I added one comment in the Review.

tests/shapiq/tests_unit/tests_game_theory/test_moebius_converter.py
```diff
@@ -22,12 +22,15 @@ def test_soum_moebius_conversion():
...
-            shapley_interactions[index] = moebius_converter(index=index, order=order)
+            shapley_interactions = moebius_converter(index=index, order=order)
             # Assert efficiency
-            assert (np.sum(shapley_interactions[index].values) - predicted_value) ** 2 < 10e-7
-            assert (shapley_interactions[index][()] - emptyset_prediction) ** 2 < 10e-7
+            assert (np.sum(shapley_interactions.values) - predicted_value) ** 2 < 10e-7
+            assert (shapley_interactions[()] - emptyset_prediction) ** 2 < 10e-7
+            for v in moebius_converter._computed.values():
+                # check that no 0's are in the interaction lookup (except for the empty set, which is the first entry)
+                interaction_lookup = v.interaction_lookup
```
mmschlk, line 32: This part here should be wrong and we should check in the `InteractionValues.interactions` dictionary, which maps the interactions onto the scores. After our latest refactor of the `InteractionValues` and `Game` objects, we kind-of removed the lookup and values attributes. Originally we split the data structure into two attributes `values` a vector containing the interaction scores and `interaction_lookup` mapping interactions (tuples) to the index of the score in the `values`. Now we combined both things into one dictionary (`interactions`) which contains the scores. 

**TLDR**: So basically we need to check weather values in  `interactions` contain zero and not the lookup mapping.

mmschlk (changes requested):

PR #480, opened 2025-11-26
Add CTEImputer 

### LINKED ISSUE #225: Could we include CTE?
CTE seems to work very [well](https://arxiv.org/abs/2406.18334v1). Thanks for that @hbaniecki. Can we include this at some point as well?

hbaniecki:
### Introducing CTEImputer 

closes https://github.com/mmschlk/shapiq/issues/225

Adds the [`CTEImputer`](https://github.com/mmschlk/shapiq/blob/main/src/shapiq/imputer/cte_imputer.py) following the [*compress then explain* (CTE)](https://openreview.net/forum?id=LiUfN9h0Lx) methodology. It replaces missing features of the explanation point by values sampled from the background data, which is first subsampled using a distribution compression algorithm, specifically [Compress++](https://openreview.net/forum?id=lzupY5zjaU9) with [Kernel Thinning](https://www.jmlr.org/papers/v25/21-1334.html). CTE has shown to provide accurate and stable estimates of explanations while being computationally efficient. It is a new default imputer in `TabularExplainer`, removing the necessity to set `sample_size`.

### TODO
- [ ] add tests
- [ ] fix Windows CI to include some missing `C++` packages

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## development 
+
+### Introducing CTEImputer [#225](https://github.com/mmschlk/shapiq/pull/225)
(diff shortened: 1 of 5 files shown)


mmschlk:
We will defer this to v2.

PR #487, opened 2026-01-07
Add explainers for nearest neighbor models

Zaphoood:
## Motivation and Context

This PR adds three new explainers for efficiently explaining sklearns `KNearestNeighbor` and `RadiusNeighborClassifier` models.

---

## Public API Changes

-   [ ] No Public API changes

---

## How Has This Been Tested?

Unittests have been added that compare the results of the explainers' efficient algorithms to baseline values which are obtained by computing the exact Shapley Values of a ground-truth Game.

---

## Checklist

---


mmschlk (changes requested):
I applied some minor changes I noticed going through the review. The biggest change is that I introduced a `ValidNNExplainerIndices = Literal["SV"]` that captures the meaning of the current KNN Explainers better (aka. only SV possible). 

There is only one outstanding change that need to be resolved before we can merge this in: We need to remove all notebooks from the repo, as I want to merge [this PR](https://github.com/mmschlk/shapiq/pull/509) before I am merging the KNN PR in. This means all notebooks need to be converted into this sphinx gallery format and examples then live under examples/<folder/plot_<name>.py. Notably, each example needs to be very quick and take not more than a couple of seconds. We will also have a way to have longer examples somewhere but currently I would like to emphasize speed here (which KNN explainers anyway are)

mmschlk (approved):

PR #515, opened 2026-04-22
fix: Tree conversion breaking when LC_NUMERIC not standard "C".

Advueu963:
## Motivation and Context

Fix Tree conversion problems due to LC_NUMERIC not set to standard C, but de_DE. As such the conversion library did not correctly parse the XGBoost/LightGBM model. 

## Public API Changes

-   [X] No Public API changes

---

## How Has This Been Tested?

We have added new tests, to check that the conversion works correclty.

## Checklist

---


mmschlk (approved):
I still do not like this hack here. Gotta think this through for v2 a bit better. But at least we now support germans. :D

PR #521, opened 2026-05-18
shapiq Benchmark

IsaH57:
## Motivation and Context
This PR consolidates and validates a full `shapiq_benchmark` layer, ensuring all benchmark types (Interventional, Pathdependent, Local XAI, Image, and TabPFN) share a consistent interface for exact value computation and game access. It also strengthens the string-based dataset/model loading flow and the known-parameter lookup path, so benchmarks can be reproduced reliably across datasets and model families while still supporting custom inputs.

## What’s included

- Benchmark protocol and ground-truth computer abstractions
- Dataset/model loading and validation, including string-based loaders and known-parameter lookups
- Benchmark implementations:
-- Interventional tree explanations in `InterventionalBench()`
-- Pathdependent tree explanations in `PathdependentBench()`
-- Local XAI explanations in `LocalXAIBench()`
-- Image explanations in `ImageBench()`
-- TabPFN explanations in `TabPFNBench()`
- Metrics for approximation quality and ranking comparison
- Configs holding optimized parameters for 6 model - dataset combinations

## Public API Changes

-   [ ] No Public API changes

---

## How Has This Been Tested?
- Manual test runs for all benchmarks
- Each test covers both string-based loading and custom model/data paths where applicable

## Checklist

---


src/shapiq_benchmark/computers.py
```diff
@@ -0,0 +1,120 @@
...
+
+        Returns:
+            InteractionValues: The computed interaction values.
+        """
+        if budget is not None:
+            _ = budget
+        self._computer.index = index
+        self._computer.max_order = order
+        return self._computer.explain_function(x=self.game.target_instance[0])
+
```
Advueu963: As we here override exact_values, I do not think we need here budget at all. If we do need it, then maybe just introducing **kwargs everywhere would resolve any issues with the previous fix. I do not like to have parameters which are then in the end not used.

src/shapiq_benchmark/computers.py
```diff
@@ -0,0 +1,120 @@
...
+        Returns:
+            InteractionValues: The computed interaction values.
+        """
+        if budget is not None:
+            _ = budget
+        return self._computer.explain_function(
+            x=self.game.x_explain,
+            index=index,
+            max_order=order,
+        )
```
Advueu963: The same here, we should remove budget from the parameters.

src/shapiq_benchmark/computers.py
```diff
@@ -0,0 +1,120 @@
...
+            index: The index for which to compute interaction values.
+            order: The order of interactions to compute.
+            budget: Optional budget for computation (not used in this computer).
+
+        Returns:
+            InteractionValues: The computed interaction values.
+        """
+        if budget is not None:
+            _ = budget
+        return self._computer(index=index, order=order)
```
Advueu963: Also here remove budget.

src/shapiq_benchmark/base.py
```diff
@@ -0,0 +1,76 @@
...
+        self.game = game
+        self._computer = ExactComputer(game=game, n_players=game.n_players, evaluate_game=False)
+
+    def exact_values(
+        self, index: IndexT, order: int, budget: int | None = None
+    ) -> InteractionValues:
+        """Compute the exact values using brute force."""
+        if budget is not None:
+            _ = budget
+        return self._computer(index=index, order=order)
```
Advueu963: Would it not be best to move the BruteForceComputer also into computers.py. Also here remove the budget parameters.

src/shapiq_benchmark/base.py
```diff
@@ -0,0 +1,76 @@
...
+    @property
+    def computer(self) -> GroundTruthComputer[T_Index_contra]:
+        """Return the ground truth computer used by the benchmark."""
+        ...
+
+    def exact_values(
+        self, index: T_Index_contra, order: int, budget: int | None = None
+    ) -> InteractionValues:
+        """Compute exact interaction values for the given index and order."""
+        ...
```
Advueu963: I think here it might be even better to use an abstract class. Because it is actually much more than only a API definition, as the methods you describe here are implemented identically across all benchmark classes. So here make it abstract class, and then you have it easier for the other benchmark classes, where mostly only the __init__ will be different

src/shapiq_benchmark/image_bench.py
```diff
@@ -0,0 +1,128 @@
...
+        self.model = model
+
+        self._game = ImageClassifier(
+            model_name=self.model,
+            n_superpixel_resnet=n_superpixel_resnet,
+            x_explain_path=image_path,
+            normalize=normalize,
+            verbose=verbose,
+        )
+        self._computer: GroundTruthComputer[IndexType] = BruteForceComputer(self._game)
```
Advueu963: You implemented an ImageComputer already in computers.py. I would then also use it here.

src/shapiq_benchmark/interventional_bench.py
```diff
@@ -0,0 +1,112 @@
...
+            raise TypeError(msg)
+
+        data_type: Literal["classification", "regression"] | None = (
+            self.dataset.data_type if self.dataset else infer_data_type(self.model)
+        )
+
+        if data_type == "regression":
+            class_index = None
+
+        x_index = 0 if x_explain is None else x_explain
```
Advueu963: Will never be None, as you set a default in the __init__.

src/shapiq_benchmark/local_xai_bench.py
```diff
@@ -0,0 +1,143 @@
...
+            data_type = infer_data_type(self.model)
+
+        if data_type == "regression":
+            class_index = None
+
+        predict_fn = _build_predict_fn(self.model, data_type, class_index)
+        x_index = 0 if x_explain is None else x_explain
+        if not isinstance(x_index, int):
+            msg = "x_explain must be an int index."
+            raise TypeError(msg)
```
Advueu963, line 112: Here seems to be a high redudancy with the InterventionalBenchmark. I would try to merge this into some common utils functions. Even better is to make InterventionalBenchmark a sub-class of LocalXAIBench, which is true also semantically.

src/shapiq_benchmark/optimization/results/lightgbm_adult_census.json
Advueu963, line 1: How where these files created? I mean what were the sweeps and can you maybe even provide the script. Because then we can easily scale this to all datasets and obtain nice defaults and plot the corresponding accuracy/r^2.

src/shapiq_benchmark/pathdependent_bench.py
Advueu963, line 1: Also here I would make it a sub-class of LocalXAI and try to remove code redundancy as much as you ccan.

src/shapiq_benchmark/setup.py
```diff
@@ -0,0 +1,273 @@
...
+}
+
+
+def register_model_builder(model_name: str, data_type: str, builder: ModelBuilder) -> None:
+    """Register a new model builder for future extensions."""
+    key = (model_name.lower(), data_type.lower())
+    if key in _MODEL_BUILDERS:
+        msg = f"Model builder already registered for {key[0]} ({key[1]})."
+        raise ValueError(msg)
+    _MODEL_BUILDERS[key] = builder
```
Advueu963: What is the purpose of this function?

src/shapiq_benchmark/setup.py
```diff
@@ -0,0 +1,273 @@
...
+    return BenchmarkDataset(
+        x_train=setup.x_train,
+        y_train=setup.y_train,
+        x_test=setup.x_test,
+        y_test=setup.y_test,
+        data_type=cast("Literal['classification', 'regression']", setup.dataset_type),
+    )
+
+
+def load_model_from_str(
```
Advueu963: I would rename into load_fit_model_from_str, or something equivalent, as otherwise the fitting is not well documented.

src/shapiq_benchmark/setup.py
```diff
@@ -0,0 +1,273 @@
...
+    model_str: str,
+    dataset: BenchmarkDataset,
+    **kwargs: object,
+) -> object:
+    """Create an unfitted model from a string identifier.
```
Advueu963: The model is fitted, so either you need to update the docstring or just rename the method and change the docstring.

src/shapiq_benchmark/setup.py
```diff
@@ -0,0 +1,273 @@
...
+def infer_data_type(model: object) -> Literal["classification", "regression"]:
+    """Infer whether a model is a classifier or regressor based on its attributes."""
+    estimator_type = getattr(model, "_estimator_type", None)
+    if estimator_type == "classifier":
+        return "classification"
+    if estimator_type == "regressor":
+        return "regression"
+    if hasattr(model, "predict_proba") or hasattr(model, "classes_"):
+        return "classification"
+    return "regression"
```
Advueu963, line 242: I recon this will only work with estimators using the sklearn api. I do not thinnk this is bad, but we should make this clear via a comment. Maybe even just enforce that the dataset/user always gives us information on what type of dataset this is. Our loading functions always should provide this. Regarding an external user, we can just enforce that he provides this information.

src/shapiq_benchmark/tabpfn_bench.py
Advueu963, line 1: Also here it is somewhat similar to the LocaXAI Benchmark. Try to make it an sub-class and reduce code redundancy

src/shapiq_benchmark/tabpfn_bench.py
```diff
@@ -0,0 +1,141 @@
...
+        )
+
+        class _TabPFNWithPredict(Protocol):
+            shapiq_predict_function: Callable[[np.ndarray], np.ndarray]
+
+        model_with_predict = cast("_TabPFNWithPredict", self.model)
+        model_with_predict.shapiq_predict_function = cast(
+            "Callable[[np.ndarray], np.ndarray]", predict_function
+        )
+
```
Advueu963, line 99: Is it for TabPFN not simply either predict or predict_proba/predict_logits? So maybe we can just directly add an attribute `.shapiq_predict_function` without needing all the casts.

src/shapiq_games/benchmark/__init__.py
Advueu963, line 1: I think I even merged the new games into shapiq already. SO might need to resolve that, before merging.

Advueu963 (changes requested):
Overall already a very nice PR.
Some things regarding the overall structure need minor changes, namely making the relationship between the benchmark classes more clear. Importantly we can reduce code redundancy, using abstract classes and inheritance of Benchmark and LocalXAIBench.

src/shapiq_benchmark/setup.py
IsaH57: I adjusted the docstring. We know this information when the dataset is given as a string. How would you expect a user to provide this info on a dataset object? As an additional string? I thought about this before and am not sure whether that is the best solution

src/shapiq_benchmark/tabpfn_bench.py
IsaH57: I am not sure about make TabPFNBench a subclass of LocalXAIBench. It eg has a different way of uploading model and dataset as it needs labels, too.

src/shapiq_benchmark/optimization/optuna_lightgbm_adult_census.py
Advueu963, line 1: Does this file contain something really different than the california one? It would be nice to maybe merge them into one file which can be called via command line args. I think this would hold for all these optimisation scripts, that you can control via command args which model to pick and which hyperparameters. Basically the model command arg, would then decide which sweep over hyperparameters is done.

src/shapiq_benchmark/optimization/optuna_lightgbm_california_housing.py
Advueu963, line 1: merge into bigger script

src/shapiq_benchmark/optimization/optuna_random_forest_adult_census.py
Advueu963, line 1: merge into bigger script

src/shapiq_benchmark/optimization/optuna_random_forest_california_housing.py
Advueu963, line 1: merge into bigger script

src/shapiq_benchmark/optimization/optuna_xgboost_adult_census.py
Advueu963, line 1: merge into bigger script

src/shapiq_benchmark/optimization/optuna_xgboost_california_housing.py
Advueu963, line 1: merge into bigger script

src/shapiq_benchmark/local_xai_bench.py
```diff
@@ -0,0 +1,167 @@
...
+    def _load_dataset_and_model(
+        self,
+        data: str | np.ndarray,
+        model: str | Model | Callable[[np.ndarray], np.ndarray],
+        *,
+        benchmark_type: str,
+        random_state: int | None,
+        class_index: int | None,
+        **kwargs: object,
+    ) -> tuple[int | None, Literal["classification", "regression"]]:
```
Advueu963, line 62: Missing a docstring, explaining that this here silently sets self.mode, self.dataset and self.data

src/shapiq_benchmark/interventional_bench.py
```diff
@@ -0,0 +1,84 @@
...
+            **kwargs: Additional keyword arguments for model building.
+        """
+        class_index, _ = self._load_dataset_and_model(
+            data,
+            model,
+            benchmark_type="interventional",
+            random_state=random_state,
+            class_index=class_index,
+            **kwargs,
+        )
```
Advueu963, line 53: I think it is somewhat dangerous to silently set here the model and data without noting this also in the method name. At least the name should suggest `_set_and_load_dataset_model(...)` with an accompaniying docstring.

src/shapiq_benchmark/pathdependent_bench.py
```diff
@@ -0,0 +1,84 @@
...
+
+        Args:
+            data: Dataset identifier (e.g. "adult_census") or a NumPy array containing the data.
+            model: Model identifier (e.g. "decision_tree") or a fitted model object.
+            x_explain: Instance to explain.
+            class_index: Class index for classification models.
+            random_state: Random state used for data split and model init.
+            **kwargs: Additional keyword arguments for model building.
+        """
+        class_index, _data_type = self._load_dataset_and_model(
```
Advueu963: Here then also the renaming.

src/shapiq_benchmark/pathdependent_bench.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @property
+    def game(self) -> TreeSHAPIQXAI:
+        """Game instance used by the Pathdependent Benchmark."""
+        return self._game
+
+    @property
+    def computer(self) -> PathdependentComputer[IndexType]:
+        """Ground truth computer used by the Pathdependent Benchmark."""
+        return self._computer
```
Advueu963: This is quite common among all the benchmark classes, So it Should maybe be already content of the base Benchmark class? because this is just redundant code at this point.

src/shapiq_benchmark/setup.py
Advueu963: So basically given the GameBenchmarkSetup we always can extract the dataset, as we control the loading itself. If the user also gives a custom dataset, we just require them also to add the type as a string. Or differently, if the target variable in the dataset (which the user must also provide in some way or another) is ordinal --> classification; numerical --> regression. But this latter approach is quite brittle.

src/shapiq_benchmark/tabpfn_bench.py
```diff
@@ -0,0 +1,133 @@
...
+    @property
+    def game(self) -> TabPFNImputer:
+        """Game instance used by the TabPFN Benchmark."""
+        return self._game
+
+    @property
+    def computer(self) -> GroundTruthComputer[IndexType]:
+        """Ground truth computer used by the TabPFN Benchmark."""
+        return self._computer
```
Advueu963: This is redundant and should be in the general class.

Advueu963 (changes requested):
Okay overall tis is a very nice refactor! Some small improvements I think would make the code even better and then it is good to go for a merge! See the comments

Advueu963 (approved):
LGTM

PR #522, opened 2026-05-20
OddSHAP approximator

Sara-ne:
## Motivation and Context

This PR adds a new OddSHAP approximator for estimating first-order Shapley values.
OddSHAP is based on the method by Fumagalli et al. (2026). The main idea is to estimate Shapley values through odd Fourier terms.

---

## Implemented
- OddSHAP-specific sampling weights
- Regression weights for the weighted least-squares problem
- Paired coalition sampling through the existing sampler
- Used InterventionalTreeExplainer to select relevant higher-order odd interactions
- Built the active odd Fourier support from:
   - the empty interaction,
   - all single-player interactions,
   - selected higher-order odd interactions
- Solved the constrained odd Fourier regression problem
- Converted the fitted odd Fourier coefficients into first-order Shapley values
- Added a ValueError when the budget is too small 

---

## Notes
This implementation doesn't add a separate ProxySPEX-style adapter. Instead, it reuses the existing tree interaction code through InterventionalTreeExplainer.
The current implementation only supports odd_only=True, because the final Shapley value computation is based on odd-cardinality Fourier terms.

---

## Public API Changes

-   [ ] No Public API changes

---

## How Has This Been Tested?

OddSHAP was verified at two levels: unit tests and a reproduction of the paper (arXiv:2602.01399). The reproduction checks our measured MSE against the values reported in the paper's Table 1 

1. Unit tests: 
tests/shapiq/tests_unit/tests_approximators/test_approximator_oddshap.py (in this PR) — 59 tests, all passing, against ExactComputer ground truth:

- Exact recovery of Shapley values at full budget
- Monotone MSE convergence as the budget grows
- Determinism under a fixed random_state
- Low-budget ValueError path (no internal fallback approximator)
- Approximator-interface conformance

2. All four estimators (MSR, SVARM; PermutationSampling, OddSHAP) reproduce the paper within a factor of ~2–3 (expected RNG/hardware / instance-set spread). 
Tabular value functions (Estate, Cancer, CG60, IL60, NHANES, Crime; d=15..101) — OddSHAP runs and converges correctly across the full dimension range. 

---

## Checklist

-   [ ] The changes have been tested locally.

---


42logos:
### Reproduction & benchmark for OddSHAP (Task 3 + Task 4)

The paper reproduction and the cross-method benchmark for this approximator live on the companion branch [`wu/oddshap-repro`](https://github.com/FabianK-Dev/shapiq/tree/wu/oddshap-repro), commit [`317b77a8`](https://github.com/FabianK-Dev/shapiq/commit/317b77a8140c5c0b5c67bd4bd4d0837181da3458) (branched off `oddshap_approximator`). It adds:

- **`examples/approximators/plot_oddshap_paper_reproduction.py`** — a gallery example reproducing the paper's Table 1/3 (Fumagalli et al., arXiv:2602.01399) on all six tabular value functions (Cancer, Estate, CG60, IL60, NHANES, Crime; `d = 15..101`).
- **`notebooks/oddshap_reproduction_and_benchmark.ipynb`** — an executed notebook covering **Task 3** (paper reproduction) and **Task 4** (cross-method benchmark on synthetic SOUM games), driving `benchmark/performance.py`.
- **`benchmark/`** — the conformance/benchmark suite (`run_sweep` over the SV approximators against `ExactComputer` ground truth).

Key points:

- **No external `shap` dependency.** The exact ground truth uses shapiq's own `InterventionalTreeExplainer(index="SV", max_order=1)`. It was cross-checked against the reference SHAP interventional `TreeExplainer` and agrees to within `1e-8` per feature on every value function, including the missing-value-heavy NHANES (using native-NaN handling, no imputation).
- **Accuracy (Task 3).** Over N = 30 local explanations per value function at a budget of ~`100·d`, **OddSHAP is rank-1 on all 6 value functions (average rank 1.00)** against MSR, SVARM, PermutationSampling, KernelSHAP, k-additive SHAP, and RegressionMSR. The four real-data value functions match the paper's reported error scale; the two synthetic ones (CG60/IL60) use shapiq's own generators, so absolute MSE differs while the ranking is unchanged.
- **Benchmark (Task 4).** On SOUM games (`n = 10`, 3 seeds) with closed-form exact Shapley values, OddSHAP's MSE collapses to numerical zero at ~1/4 of the exact budget, ahead of all sampling baselines.

A 7-page external report with the full configuration, Table 1 (median + IQR), and all figures accompanies this (shared by email).

Sara-ne:
Hi @mmschlk this PR is ready for review now

src/shapiq/approximator/regression/oddshap.py
```diff
@@ -0,0 +1,575 @@
...
+def _import_lightgbm() -> ModuleType:
+    try:
+        lightgbm = import_module("lightgbm")
+    except ImportError as err:
+        msg = (
+            "The 'lightgbm' package is required for OddSHAP but it is not installed. "
+            "Install it via the optional extra: pip install 'shapiq[proxy]'."
+        )
+        raise ImportError(msg) from err
+    return lightgbm
```
Advueu963: We should probably do this on the import in `__init__.py` in approximator to guard against the user not having lightgbm. What you can do is to create a dummy class as we have done on main.

src/shapiq/approximator/regression/oddshap.py
```diff
@@ -0,0 +1,575 @@
...
+
+class OddSHAP(Approximator):
+    """OddSHAP approximator for first-order Shapley values (Fumagalli et al., 2026).
+
+    Note:
+        Where Algorithm 1 of the paper falls back to TreeSHAP for budgets below
+        ``n * interaction_factor``, this implementation raises ``ValueError`` instead
+        (no silent downgrade to another estimator), unless the budget already covers
+        the full coalition space (``budget >= 2**n``). It therefore does not reproduce
+        the low-budget, high-dimension regime of the paper's Figure 2.
```
Advueu963, line 117: Why is that the case? You can use the implemented InterventionalTreeExplainer to extract the true values of the tree fitted on the binary data.

src/shapiq/approximator/regression/oddshap.py
```diff
@@ -0,0 +1,575 @@
...
+            weight_vector[coalition_size] = 1.0 / (
+                coalition_size * (n - coalition_size) * binom(n, coalition_size)
+            )
+        return weight_vector
+
+    def __init__(
+        self,
+        n: int,
+        *,
+        pairing_trick: bool = True,
```
Advueu963: I would not leave the option of pairing_trick for ODDSHAP, as this is necessary for it to work efficiently.

src/shapiq/approximator/regression/oddshap.py
```diff
@@ -0,0 +1,575 @@
...
+        # covers the full coalition space is always sufficient, even when 2**n is
+        # smaller than the eta-based minimum (small n).
+        minimum_budget = min(self.n * self.interaction_factor, 2**self.n)
+        if budget < minimum_budget:
+            msg = (
+                "The budget is too small for OddSHAP. "
+                f"Received budget={budget}, but at least {minimum_budget} evaluations are required. "
+                "Please increase the budget."
+            )
+            raise ValueError(msg)
```
Advueu963, line 236: Here you should then integrate the InterventionalTreeExplainer.

src/shapiq/approximator/regression/oddshap.py
```diff
@@ -0,0 +1,575 @@
...
+        params = {
+            "verbose": -1,
+            "n_jobs": 1,
+            "random_state": self._random_state,
+            "max_depth": 10,
+            **(self.tree_params or {}),
+        }
+        surrogate_model = lgb.LGBMRegressor(**params)
+        surrogate_model.fit(coalitions.astype(float), game_values)
+        return surrogate_model
```
Advueu963: In the paper they actually used a GridSearch informed lightgbm model. So to reproduce the papers result we should as default for lightgbm provide these GridSearch parameters. You can find them on the main branch within the refactored ProxySPEX module (_model.py will contain them)

src/shapiq/approximator/regression/oddshap.py
```diff
@@ -0,0 +1,575 @@
...
+
+        tree_models = convert_tree_model(surrogate_model)
+        # The ProxySPEX instance is only a host for its exact GBT->Fourier extractor;
+        # it is never sampled or fit.
+        # max_order=1 keeps the base class from building an order-2 interaction
+        # lookup that the extractor never uses.
+        fourier_extractor = ProxySPEX(
+            n=self.n, max_order=1, proxy_model=surrogate_model, random_state=self._random_state
+        )
+        fourier_coefficients = fourier_extractor._sklearn_to_fourier(tree_models)  # noqa: SLF001
```
Advueu963: Ahh nice! So here we could do even a little bit more refactoring regarding import stability of ODDSHAP. So yes in the paper they have choosen a lightgbm, but here we can actually use any arbitrary tree supported by the converter. So what I would find quite nice, is to have OddSHAP if lightgbm is not installed simply use a DecisionTreeRegressor alongside a warning. We have introduced this behaviour in the refactored ProxySHAP/ProxySPEX branches (_models.py). Basically if the optimal configuration is not loaded they we warn the user. So here OddSHAP could also use similar structure with a proxy_model parameter to determine which tree model should be used. You can orientate yourself at ProxySHAP/ProxySPEX

src/shapiq/approximator/regression/oddshap.py
```diff
@@ -0,0 +1,575 @@
...
+                normalized = tuple(sorted(interaction))
+
+                # skip empty or singleton terms because OddSHAP always includes them
+                if len(normalized) <= 1:
+                    continue
+
+                # keep only odd-sized higher-order interactions
+                if len(normalized) % 2 == 0:
+                    continue
+
```
Advueu963: Is this branch even reachable? Becuase the normalized interactions in the selected_interactions can only be odd, based on the method as you already filter out the even ones in `_select_odd_interactions`. Or did I miss something here?

src/shapiq/approximator/regression/oddshap.py
```diff
@@ -0,0 +1,575 @@
...
+        sv_values[0] = baseline_value
+
+        for interaction, position in self.odd_interaction_lookup.items():
+            # The empty interaction does not contribute to player attributions
+            if len(interaction) == 0:
+                continue
+
+            # OddSHAP only uses odd-cardinality Fourier terms for the attribution
+            if len(interaction) % 2 == 0:
+                continue
```
Advueu963: Is this not also guaranteed, as the interactions the odd fourier refines are only on odd fourier interactions?

src/shapiq/approximator/regression/oddshap.py
```diff
@@ -0,0 +1,575 @@
...
+
+            coefficient = odd_fourier_coefficients[position]
+            share = coefficient / len(interaction)
+
+            for player in interaction:
+                sv_values[player + 1] += share
+
```
Advueu963: Actually I think you can write this efficiently using numpy vectorization. So it would be possible for each player, to write an interaction vector where the player is contained, multiply by the vector of odd interactions and then divide by the vector with the sizes of the interactions. I think this would be speedup especially for larger games.

src/shapiq/approximator/regression/oddshap.py
```diff
@@ -0,0 +1,575 @@
...
+            max_order=1,
+            min_order=0,
+            n_players=self.n,
+            # the base class populated {(): 0, (0,): 1, ..., (n-1,): n} for
+            # min_order=0, max_order=1
+            interaction_lookup=self.interaction_lookup,
+            baseline_value=float(empty_set_value),
+            estimated=not (budget >= 2**self.n),
+            estimation_budget=budget,
+            target_index=self.index,
```
Advueu963: This is not necessary for "SV". Simply `index=self.index` is enough.

src/shapiq/approximator/regression/oddshap.py
```diff
@@ -0,0 +1,575 @@
...
+        **kwargs: Any,
+    ) -> None:
+        """Initialize the OddSHAP approximator.
+
+        ``tree_params`` entries override the surrogate defaults — including
+        ``random_state``, ``n_jobs``, and ``verbose``; ``max_depth`` defaults to 10
+        (the paper's configuration) unless overridden.
+        """
+        del kwargs
+        _import_lightgbm()
```
Advueu963: I think we can get rid of this. See the comments on the lightgbm->fourier coefficient extraction.

src/shapiq/approximator/regression/oddshap.py
```diff
@@ -0,0 +1,575 @@
...
+        return weight_vector
+
+    def __init__(
+        self,
+        n: int,
+        *,
+        pairing_trick: bool = True,
+        sampling_weights: np.ndarray | None = None,
+        random_state: int | None = None,
+        odd_only: bool = True,
```
Advueu963: What does the odd_only component mean? Is OddSHAP not designed to be only fitted on the odd interactions?

tests/shapiq/tests_unit/tests_approximators/test_approximator_oddshap.py
```diff
@@ -0,0 +1,740 @@
...
+import shapiq.approximator as approximator_module
+from shapiq.approximator.regression import OddSHAP
+from shapiq.game_theory.exact import ExactComputer
+from shapiq.interaction_values import InteractionValues
+from shapiq_games.synthetic import SOUM, DummyGame
+from tests.shapiq.markers import skip_if_no_lightgbm
+
+# OddSHAP's surrogate requires lightgbm (optional 'proxy' extra); skip the whole
+# module in environments without it, like the other LightGBM-dependent suites.
+pytestmark = [skip_if_no_lightgbm]
```
Advueu963: Yeah we might not need this, if we adjust the module oddshap.py to work per-default with a decisionTree regressor in the case lightgbm is not installed.

Advueu963 (changes requested):
Overall a very nice implementation! I would work on some redudant code regarding guarding against odd interactions. The way I see it , ODDSHAP is designed to work/exxtract only odd components, so I would not leave the options to also get non-odd options. The same holds for activating/deactivating the pariing trick. Also I would like to improve the robustness of the implementation by using the same tricks done in ProxySHAP and the implementations in `_models.py`. Also in that move it might be nice to move OddSHAP to the proxy module, as it is basically using a proxy model to extract the fourier interactions.

src/shapiq/approximator/regression/oddshap.py
42logos: Good point — re-reading the paper, Algorithm 1 (p. 6) does prescribe if m < d*eta: return phi via TreeSHAP on the fitted GBT, and the reference implementation uses InterventionalTreeExplainer for exactly this.

However, @mmschlk previously gave us the guidance that OddSHAP should not silently fall back to a different estimator when the budget is insufficient — raise an error instead, so the user knows exactly what happened. That's why the current implementation raises ValueError here and the docstring documents it as a deliberate deviation from Algorithm 1.

We're happy to implement the paper's TreeSHAP-on-surrogate fallback if that's the agreed direction — just want to make sure this aligns with Max's earlier guidance first. The two positions are:
 - Paper: low-budget → run TreeSHAP on the already-fitted GBT surrogate, return those Shapley values
 - Current: low-budget → raise ValueError, user must either increase the budget or explicitly use another estimator

@mmschlk could you weigh in on which you'd prefer for the library?

src/shapiq/approximator/regression/oddshap.py
42logos: I checked this against the paper — Section 5 (p. 7) states: "All proxy-based methods, including ProxySPEX used by OddSHAP, use the same LightGBM proxy with default configurations and maximum depth set to 10." The reference implementation also has grid_search=False as the default.

The GridSearch parameters in _models.py (max_depth: [3, 5], max_iter: [500, 1000]) are tuned for the ProxySHAP/ProxySPEX pipeline (Butler et al., 2025) and use different depth ranges (3-5 vs. the paper's 10). Using them for OddSHAP would deviate from the paper's experimental configuration rather than reproduce it.

 We'll keep LightGBM with max_depth=10 and default params as the paper prescribes. Happy to add an optional hpo=True path in a follow-up if there's interest, but the default should match the paper.

src/shapiq/approximator/regression/oddshap.py
42logos: this is the same point as comment #2 — whether to implement Algorithm 1's if m < d*eta: return phi via TreeSHAP branch using InterventionalTreeExplainer(index="SV") on the surrogate.

src/shapiq/approximator/regression/oddshap.py
Sara-ne: I agree with @42logos here. This was also how I understood the situation.

I had actually implemented the fallback before, using the existing `InterventionalTreeExplainer`, but then changed it to raise a `ValueError` based on the earlier feedback that OddSHAP shouldn't silently switch to another estimator when the budget is too low.

So the current behavior is intentional, but it is a deliberate deviation from Algorithm 1 of the paper.

If the preferred direction is to follow the paper more closely, we can switch back to the older implementation for the low-budget case.

src/shapiq/approximator/regression/oddshap.py
42logos: the authors' own reference implementation (FFmgll/oddshap (https://github.com/FFmgll/oddshap)) defaults to grid_search=False.

src/shapiq/approximator/regression/oddshap.py
42logos: yes  this branch is unreachable from approximate() because _select_odd_interactions already filters to len >= 3 and len % 2 == 1. We've kept it as a defensive guard since _build_support is a public method that could be  called directly with arbitrary input, but your analysis is right that the normal code path never hits it. what do you think about it?

Advueu963: I do not have the feeling that this method is such a public method, as it can only be called with an instantiated object, and also uses `_` in the beginning, which is often an indicator for a private method. Also, I do not see the necessity of this as a public method, as no approximator currently requires it. I would thus leave this out, but you can, of course, have a different opinion. In line with this public/private method discussion, we should probably refactor also the way of extracting the Fourier interactions, as this now also does this object instantiation and access a private method ([see comment](https://github.com/mmschlk/shapiq/pull/522#discussion_r3408503017)).

src/shapiq/approximator/regression/oddshap.py
Advueu963: Thank you for pointing this out! Then we should probably just stick to that what the paper had used as default.

src/shapiq/approximator/regression/oddshap.py
Advueu963: Ahh, I see. I was not aware of this internal discussion. ^^ Okay, then I'd say we leave the decision to @mmschlk. But we should be aware that reproducing the paper is quite difficult with this decision, as the approximator will error when the budget is not sufficient. Therefore, I would argue that it would be better to incorporate the fallback, as it enables 1:1 replication of the original paper's experiments. But of course you guys can decide otherwise :D.

src/shapiq/approximator/regression/oddshap.py
42logos: Done — OddSHAP now falls back to a DecisionTreeRegressor with a UserWarning when LightGBM is not installed, following the ProxySHAP/ProxySPEX resolution pattern. The old hard ImportError is gone.

src/shapiq/approximator/regression/oddshap.py
42logos: OddSHAP now tries LightGBM first and falls back to DecisionTreeRegressor with a UserWarning (including listing any dropped LightGBM-specific tree_params), following the ProxySHAP/ProxySPEX pattern. We also had to add a list-wrapping guard in _select_odd_interactions because convert_tree_model(DecisionTreeRegressor(...)) returns a single TreeModel rather than a one-element list.

src/shapiq/approximator/regression/oddshap.py
42logos: Yes, guaranteed — the active support only contains the empty set and odd-cardinality interactions. Removed the explicit check by vectorizing the entire transform

src/shapiq/approximator/regression/oddshap.py
42logos: Done — replaced by _resolve_surrogate_model for the LightGBM/DecisionTree resolution, and we also decoupled the Fourier extraction entirely from ProxySPEX

src/shapiq/approximator/regression/oddshap.py
42logos: Ahh yes. Our reproduction actually does cover the full budget range including below n * eta — the budget curve script (plot_oddshap_budget_curves.py (https://github.com/FabianK-Dev/shapiq/blob/wu/oddshap-repro/examples/approximators/plot_oddshap_budget_curves.py), (https://github.com/FabianK-Dev/shapiq/blob/wu/oddshap-repro/notebooks/oddshap_reproduction_and_benchmark.ipynb)) sweeps m from d + 1 to min(2^d, 20000). When OddSHAP hits the budget threshold it raises ValueError, which the script catches and simply skips that data point — so the OddSHAP curve starts at n * eta while the baseline curves extend to lower budgets, which matches the paper's Figure 2 behavior. The executed reproduction notebook is here (https://github.com/FabianK-Dev/shapiq/blob/wu/oddshap-repro/notebooks/oddshap_reproduction_and_benchmark.ipynb) and the cluster results are in notebooks/cluster_results/ (https://github.com/FabianK-Dev/shapiq/tree/wu/oddshap-repro/notebooks/cluster_results).

src/shapiq/approximator/regression/oddshap.py
42logos: Done — _transform_to_shapley is now a single vectorized operation: sv_values[1:] = -2.0 * (masks.T @ (coeffs / sizes)). Eliminates the per-player Python loop entirely.

src/shapiq/approximator/regression/oddshap.py
42logos: Changed to index=self.index and dropped target_index.

tests/shapiq/tests_unit/tests_approximators/test_approximator_oddshap.py
42logos: Done — the module-level skip_if_no_lightgbm marker is removed. Three tests that assert LightGBM-specific params keep individual @skip_if_no_lightgbm decorators since those params don't exist on DecisionTreeRegressor.

mmschlk (approved):
This can now be merged. I think the decision to raise an error instead of silently falling back to another approximation algorithm is the more honest and correct handling of this case.

PR #519, opened 2026-05-15
Refactoring TreeExplainer, Adding InterventionalTreeExplainer and RegressionMSR approximator.

Advueu963:
## Motivation and Context

We refactor the TreeExplainer to incorporate and support both LinearTreeSHAP and an interventional mode. As such, we also add the InterventionalTreeExplainer. We refactor the tree conversion to use a custom C++ parser, resulting in at least 2x faster conversion than shap. 
tree_conversion_benchmark

We also add [RegressionMSR](https://neurips.cc/virtual/2025/loc/san-diego/poster/118119)  and ProxySHAP two novel model-agnostic approximation methods for cardinal-probabilistic values (SV,BV) and cardinal-probabilistic interactions (SII,BII,FSII,FBII), respectively.

## Public API Changes

-   [ ] No Public API changes
-   [X] Yes, Public API changes (Details below)

---

## How Has This Been Tested?

Tests have been added and also been locally tested.

## Checklist

-   [X] The changes have been tested locally.
-   [X] Documentation has been updated (if the public API or usage changes).
-   [X] An entry has been added to `CHANGELOG.md` (if relevant for users).
-   [X] The code follows the project's style guidelines.
-   [X] I have considered the impact of these changes on the public API.

PR #532, opened 2026-05-27
fix(InteractionValues): skip dense lookup build for sparse-dict input

### LINKED ISSUE #531: InteractionValues.__init__ allocates dense interaction lookup for sparse dict input -> OOM at n_players >= ~25
# `InteractionValues.__init__` allocates dense interaction lookup for sparse `dict` input → OOM at `n_players` ≥ ~25

## Summary

`shapiq.InteractionValues.__init__` builds a dense interaction lookup of size
`Σ_{k=min_order..max_order} C(n_players, k)` whenever the caller does not pass
`interaction_lookup`, **including when `values` is already a sparse
`dict[tuple[int, ...], float]`**. In the dict branch the built lookup is then
discarded — it is neither read by `_validate_and_return_interactions`, nor
returned to the caller, nor stored on the resulting instance. For
`n_players >= ~25` with `max_order` not capped, this OOMs.

## Minimal reproduction

```python
from shapiq.interaction_values import InteractionValues

iv = InteractionValues(
    values={(0,): 1.0, (1,): 2.0, (0, 1): 3.0},  # 3-entry sparse dict
    index="SV",
    max_order=30,
    n_players=30,
    min_order=1,
    baseline_value=0.0,
)
# MemoryError at interaction_values.py:1194 ->
#   generate_interaction_lookup(players=30, min_order=1, max_order=30)
#   → tries to enumerate ~2**30 ≈ 1e9 interaction tuples
```

Tested against current `main` (commit `57c254b0`).

## Where the bug lives

`src/shapiq/interaction_values.py`, `_validate_and_return_interactions`
(around lines 1192–1208):

```python
interactions: dict[tuple[int, ...], float] = {}
if interaction_lookup is None:                          # <-- unconditional
    interaction_lookup = generate_interaction_lookup(   # <-- O(2^n_players) worst case
        players=n_players, min_order=min_order, max_order=max_order,
    )
# ... type check ...
if isinstance(values, dict):
    interactions = copy.deepcopy(values)                # <-- lookup is NOT used here
else:
    interactions = {
        interaction: values[index].item()
        for interaction, index in interaction_lookup.items()   # <-- lookup is only used here
    }
```

The built `interaction_lookup` is:

- **Not read** in the `isinstance(values, dict)` branch — that branch only does
  `copy.deepcopy(values)`.
- **Not returned** from the function — the return type is
  `dict[tuple[int, ...], float]` (just `interactions`).
- **Not stored** on the resulting `InteractionValues` instance — the class
  exposes `interaction_lookup` as a `@property` derived from
  `self.interactions.items()` (lines ~150–160), not as a stored attribute.

So in the dict branch the lookup is purely a wasted allocation.

## How it surfaces in practice

`InterventionalTreeExplainer.explain_function` (post-#519, since the sparse
routing was added) wraps its sparse result via:

```python
# src/shapiq/tree/interventional/explainer.py:498
return InteractionValues(
    interactions,            # sparse dict produced by compute_interactions_batched_sparse
    max_order=self.max_order,
    min_order=1,
    index=computation_index,
    n_players=self.n_players,
    baseline_value=self.baseline_value,
    target_index=self.index,
    # no interaction_lookup passed
)
```

With `max_order = n_players` (e.g., 30 for the Breast Cancer dataset), this
triggers the OOM above immediately — even though the sparse computation itself
finishes happily with only ~1024 entries.

## Expected behavior

Constructing `InteractionValues` from a sparse `dict` should not allocate a
dense lookup whose size depends on `2^n_players`. The dict already carries
its own interaction tuples as keys; the lookup is only required to map
ndarray indices back to interactions, which is the other input form.

## Suggested fix

Only build the dense lookup in the branch that actually consumes it (the
`np.ndarray` branch). A 1-line guard suffices:

```diff
-    if interaction_lookup is None:
+    if interaction_lookup is None and not isinstance(values, dict):
         interaction_lookup = generate_interaction_lookup(
             players=n_players, min_order=min_order, max_order=max_order,
         )
```

A PR with this fix + a regression test will be opened shortly.

## Notes on alternative readings I considered

I considered whether the design intent might be "sparse dict input *requires*
an explicit `interaction_lookup`", which would re-interpret the OOM as user
error rather than a bug. Several pieces of evidence rule this out:

1. The `interaction_lookup` parameter is documented as defaulting to `None`,
   which the docstring says "will be generated from the n_players, min_order,
   and max_order parameters" — i.e., `None` is explicitly a supported value.
2. The `isinstance(values, dict)` branch never reads `interaction_lookup` — if
   the pairing were a contract, you'd expect at least
   `values.keys() ⊆ interaction_lookup.keys()` validation.
3. `InteractionValues.interaction_lookup` is a `@property` derived from
   `self.interactions.items()` rather than a stored field — the class treats
   the dict as the canonical source of truth for which interactions exist.
4. The caller in `InterventionalTreeExplainer.explain_function` (cited above)
   does pass a sparse dict without a lookup; that code is part of the post-#519
   sparse refactor and clearly intends this pattern to work.

## Environment

- shapiq commit: `57c254b0` (`main`, 2026-05-26)
- Python 3.14.2, Windows 11
- No special build flags; uses the standard wheel

42logos:
# fix(InteractionValues): skip dense lookup build for sparse-dict input

Closes #531

## Problem

`_validate_and_return_interactions` (in `src/shapiq/interaction_values.py`)
unconditionally calls `generate_interaction_lookup(n_players, min_order,
max_order)` when `interaction_lookup` is `None` — including when `values` is
a sparse `dict[tuple[int, ...], float]`. In the dict branch, the built lookup
is then discarded:

- The `isinstance(values, dict)` branch only does `copy.deepcopy(values)` and
  does not reference `interaction_lookup`.
- The function returns only the `interactions` dict; the lookup is not
  passed back to the caller.
- `InteractionValues.interaction_lookup` is a `@property` derived from
  `self.interactions.items()`, not a stored field — the built lookup is not
  retained on the instance either.

For `n_players >= ~25` with `max_order` not capped, the lookup has up to
`2**n_players` entries and `MemoryError`s.

This is observable on the post-PR-#519 sparse path of
`InterventionalTreeExplainer.explain_function`, which wraps a sparse result
via `InteractionValues(values=dict, interaction_lookup=None, ...)` and
triggers the OOM on any tree-based value function with `d >= ~25`.

See linked issue for the minimal reproduction and the analysis of why this is
a bug rather than an "implicit must-pair-lookup" API contract.

## Fix

One-line guard: only build the dense lookup when it will actually be consumed
(i.e., when `values` is an `ndarray`).

```diff
-    if interaction_lookup is None:
+    if interaction_lookup is None and not isinstance(values, dict):
         interaction_lookup = generate_interaction_lookup(
             players=n_players, min_order=min_order, max_order=max_order,
         )
```

Behavior in all four input combinations is preserved:

| `values` | `interaction_lookup` | Pre-fix | Post-fix |
|---|---|---|---|
| `dict`    | `None`        | Build (then discard) → OOM at large `n` | Skip build, `deepcopy(values)` |
| `dict`    | caller-supplied | Type-check, then `deepcopy(values)` | Unchanged |
| `ndarray` | `None`        | Build, then index by `lookup.items()` | Build (gated), then index — same |
| `ndarray` | caller-supplied | Type-check, then index | Unchanged |

The type check `if interaction_lookup is not None and not isinstance(...)` on
the following line still validates caller-supplied lookups in all four cases.

## Tests

Adds `tests/shapiq/tests_unit/test_interaction_values.py::
test_init_with_sparse_dict_does_not_build_dense_lookup` — a regression test
constructing `InteractionValues(values={(0,): 1.0, (1,): 2.0, (0, 1): 3.0},
n_players=30, max_order=30, ...)`. This `MemoryError`s pre-fix and completes
in ~0.13 s post-fix.

Full test file passes (`47 passed`) on the changed branch.

## Test plan

- [x] Regression test added and passing
- [x] Full `tests/shapiq/tests_unit/test_interaction_values.py` passes
- [ ] CI green (will verify after push)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,7 +102,7 @@ The conversion of the tree methods has been moved to C++ giving at least 2x up t
 - fixes `min_order` in `TreeExplainer` so that it now actually restricts the returned `InteractionValues` to interactions of order ``min_order..max_order`` (``min_order=0`` continues to include the empty interaction at the baseline value); invalid values now raise a clear `ValueError`. [#325](https://github.com/mmschlk/shapiq/issues/325)
 - fixes tree conversion breaking when the `LC_NUMERIC` locale is not set to the standard `"C"` value. [#515](https://github.com/mmschlk/shapiq/pull/515)
 - fixes a segfault in the `ProxySHAP` C++ extension code. [#506](https://github.com/mmschlk/shapiq/pull/506)
-
+- fixes a Out Of Memory (OOM) when values is an dictionary and n_players is large.  [#531](https://github.com/mmschlk/shapiq/issues/531)
 
 ## v1.4.1 (2025-11-10)
 
--- a/src/shapiq/interaction_values.py
+++ b/src/shapiq/interaction_values.py
@@ -1189,22 +1189,20 @@ def _validate_and_return_interactions(
     Raises:
         TypeError: If the values or interaction_lookup are not of the expected types.
     """
-    interactions: dict[tuple[int, ...], float] = {}
-    # Dict-input does not consume the lookup (see deepcopy branch below) and
-    # building it is O(2**n_players) worst case — skip to avoid OOM at large n.
-    if interaction_lookup is None and not isinstance(values, dict):
-        interaction_lookup = generate_interaction_lookup(
-            players=n_players,
-            min_order=min_order,
-            max_order=max_order,
-        )
-    if interaction_lookup is not None and not isinstance(interaction_lookup, dict):
-        msg = f"Interaction lookup must be a dictionary. Got {type(interaction_lookup)}."
-        raise TypeError(msg)
-
+    interactions: dict[tuple[int, ...], float]
     if isinstance(values, dict):
-        interactions = copy.deepcopy(values)  # type: ignore[assignment]
+        # Shallow copy is sufficient, as keys and values are immutable (tuples[float,...] and float).
+        interactions = dict(values)
     else:
+        if interaction_lookup is None:
+            interaction_lookup = generate_interaction_lookup(
+                players=n_players,
+                min_order=min_order,
+                max_order=max_order,
+            )
+        if not isinstance(interaction_lookup, dict):
+            msg = f"Interaction lookup must be a dictionary. Got {type(interaction_lookup)}."
+            raise TypeError(msg)
         interactions = {
             interaction: values[index].item() for interaction, index in interaction_lookup.items()
         }
--- a/src/shapiq/interaction_values.py
+++ b/src/shapiq/interaction_values.py
@@ -1190,7 +1190,9 @@ def _validate_and_return_interactions(
         TypeError: If the values or interaction_lookup are not of the expected types.
     """
     interactions: dict[tuple[int, ...], float] = {}
-    if interaction_lookup is None:
+    # Dict-input does not consume the lookup (see deepcopy branch below) and
+    # building it is O(2**n_players) worst case — skip to avoid OOM at large n.
+    if interaction_lookup is None and not isinstance(values, dict):
         interaction_lookup = generate_interaction_lookup(
             players=n_players,
             min_order=min_order,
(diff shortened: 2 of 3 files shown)


Advueu963:
Thank you for your help!

I have further modified the code, as the current deepcopy was unnecessary and only increased the runtime.

Advueu963 (approved):
PR is fine and improves the scalability.

PR #539, opened 2026-06-04
Improving ProxySHAP, additional Approximators for TabularExplainer, increasing shapiq robustness against huge feature counts

### LINKED ISSUE #538: Make proxy-based Approximators available through Explainers API
### Describe the workflow you want to enable

Make it possible for users to easily choose (ProxySHAP, RegressionMSR, and ProxySPEX) at the Explainer level.

### Describe your proposed solution

Add a string type definition for "proxy".

### Describe alternatives you've considered, if relevant

_No response_

### Additional context

Proxy-based approximators are sota in most instances and should be available for the public facing Explainer API.

### Impact

None

Advueu963:
## Motivation and Context

This PR first improves ProxySHAP by adding support for linear models and HPO variants using sklearn.
Additionally, we add the capability for ProxySHAP and ProxySPEX as approximators for the TabularExplainer.

Most importantly, we extend the number of features that correctly work in the Shapiq environment via log-based weight computation in `CoalitionSampler` and approximators, beyond 1000 features.

Also, smaller bugfixes are implemented, which lead to suboptimal performance of ProxySHAP and RegressionMSR because `random_state` is not correctly aligned between the sampled coalitions of the tree and the adjustment method.

Closes: #538 

## Public API Changes

-   [ ] No Public API changes

## How Has This Been Tested?

Benchmarking the implementations against datasets and reproducing the results of the RegressionMSR and ProxySHAP paper.

## Checklist


CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
 # Changelog
 
+## v1.5.2 (unknown)
+
+### Highlights of new Features
+
+- Added `'baseline'` imputer support (`BaselineImputer`) to the `shapiq_games` `LocalExplanation` benchmark games.
+- Added `ProxySHAP` and `ProxySPEX` as approximator options for the `TabularExplainer` and `TabPFNExplainer` in `shapiq.explainer`.
```
mmschlk: Why only to those explainers and not agnostic/ bare explainers?

pyproject.toml
```diff
@@ -70,6 +70,8 @@ proxy = [
     # required by shapiq.approximator.proxy (ProxySHAP, ProxySPEX, RegressionMSR)
     "xgboost",   # default proxy model in ProxySHAP
     "lightgbm",  # required by ProxySPEX
+    "catboost",
+    "smac", # SMAC HPO for proxy models
```
mmschlk, line 74: Not sure if this is here to stay! But for now it is definitely great to have it in the optional group!

src/shapiq/approximator/proxy/_models.py
```diff
@@ -0,0 +1,380 @@
...
+            msg = "XGBoost is not installed. Install it with: pip install 'shapiq[proxy]' or choose a different proxy_model."
+            warn(msg, stacklevel=2)
+        if proxy_str == "lightgbm":
+            try:
+                from lightgbm import LGBMRegressor
+
+                return LGBMRegressor(random_state=random_state)
+            except ImportError:
+                msg = "LightGBM is not installed. Install it with: pip install 'shapiq[proxy]' or choose a different proxy_model."
+                warn(msg, stacklevel=2)
```
mmschlk: I think this is not correct here right? This code should not be reachable, no? If proxy string was XGBoost it cannot be lightgbm right after. So the fallback of returning lgbm if available will not happen in the case it is installed and XGBoost is not. Or am I missing something obvious here?

src/shapiq/approximator/proxy/_models.py
```diff
@@ -0,0 +1,380 @@
...
+            warn(msg, stacklevel=2)
+
+        if proxy_str == "xgboost":
+            try:
+                from xgboost import XGBRegressor
+
+                return XGBRegressor(random_state=random_state)
+            except ImportError:
+                msg = "XGBoost is not installed. Install it with: pip install 'shapiq[proxy]' or choose a different proxy_model."
+                warn(msg, stacklevel=2)
```
mmschlk: Same as above

src/shapiq/approximator/proxy/_models.py
```diff
@@ -0,0 +1,380 @@
...
+        else:
+            msg = f"Unsupported SMAC proxy estimator: {name}"
+            raise ValueError(msg)
+
+    def predict(self, X: np.ndarray) -> np.ndarray:
+        """Predict using the best estimator found by SMAC."""
+        return self.best_estimator_.predict(X)
+
+
+def _select_base_proxy_via_string(proxy_str: str, random_state: int | None) -> ProxyModel:
```
mmschlk: I would like to refrain from using bare str in places that are options like this. I would like a literal here. With the four available options (counting tree as well and linear)

src/shapiq/approximator/proxy/_models.py
```diff
@@ -0,0 +1,380 @@
...
+                msg = "XGBoost is not installed. Install it with: pip install 'shapiq[proxy]' or choose a different proxy_model."
+                warn(msg, stacklevel=2)
+
+    # Tree route: We try a Decision Tree regressor, which is the fallback if no other tree model is available.
+    if proxy_str != "linear":
+        msg = f"Proxy model '{proxy_str}' is not available. Falling back to DecisionTreeRegressor."
+        warn(msg, stacklevel=2)
+        return DecisionTreeRegressor(random_state=random_state)
+    # Linear route: We use a simple linear regression model, which is the only option for a linear proxy.
+    return LinearRegression()
```
mmschlk: Could this also be Linear + Interactkons such as in the paper? For later.

mmschlk: Found the Polynomial Features myself. 🫠🚀

src/shapiq/approximator/proxy/proxyshap.py
```diff
@@ -5,25 +5,255 @@
...
+
+    # 3. Optional adjustment of the proxy. The adjustment approximator re-samples the same
+    # coalitions (ProxySHAP fixes a shared random_state), so the residual values stay aligned.
+    if adjustment_method is not None:
+        residual_values = coalition_values - fitted.predict(coalitions_matrix)
+        residual_values -= residual_values[0]  # Normalize residuals
+        residual_game = ResidualGame(n_players=n_players, game_values=residual_values)
+        proxy_interactions += adjustment_method.approximate(budget, residual_game)
+    proxy_interactions.baseline_value = baseline_value
+    proxy_interactions.interactions[()] = baseline_value  # Ensure empty coalition value is correct
```
mmschlk: These lines are duplicated across the different routes and should be refactored.

src/shapiq/approximator/proxy/proxyshap.py
```diff
@@ -5,25 +5,255 @@
...
+    proxy_interactions = InteractionValues(
+        values=proxy_values.interactions,
+        index=approximation_index,
+        max_order=max_order,
+        n_players=n_players,
+        min_order=0,
+        estimated=budget >= 2**n_players,
+        estimation_budget=budget,
+        baseline_value=float(baseline_value),
+        target_index=target_index,
```
mmschlk: Same comment here, these lines seem to be duplicated a lot. We could move them inside the dispatch function and only dispatch on a private helper function and collect the date results in one place.

src/shapiq/approximator/proxy/proxyspex.py
```diff
@@ -26,6 +33,13 @@
 
 ValidProxySPEXIndices = ValidMoebiusConverterIndices
 
+# Hyperparameter grid for ProxySPEX's reference HPO-informed LightGBM proxy (Butler et al., 2025).
+_LIGHTGBM_DECODER_GRID = {
+    "max_depth": [3, 5],
+    "max_iter": [500, 1000],
+    "learning_rate": [0.01, 0.1],
+}
```
mmschlk: Nice! I might be blind but I would love to have such a config also in ProxySHAP, if possible? Potentially we should move these defaults actually inside the _models module and import them from there. I feel like this is the more natural space for it as we do not really have a hard default here and users are pretty welcome to change.

tests/shapiq/tests_unit/tests_approximators/test_approximator_proxyspex.py
```diff
@@ -88,14 +88,21 @@ def dummy_game(X):
         assert estimates[interaction] == pytest.approx(1.0, abs=0.5)
 
 
-def test_proxyspex_import_error_if_missing(monkeypatch):
-    """Tests that ProxySPEX raises an ImportError if lightgbm is not installed."""
+def test_proxyspex_falls_back_to_decision_tree_without_lightgbm(monkeypatch):
```
mmschlk: It would probably be nice to have tests that show that ProxySPEX and proxyshap will use lightgbm/XGBoost if the other was requested but not installed. This however is probably not possible currently since the case is not hit! Haha 😜

src/shapiq/utils/sets.py
```diff
@@ -28,6 +29,36 @@
 T = TypeVar("T", int, str)
 
 
+def log_binom(n: int, k: int | np.ndarray) -> float | np.ndarray:
```
mmschlk, line 32: I did not find explicit tests for this utils function. Maybe I missed it.

CHANGELOG.md
mmschlk: Yes I see this is not properly wired then, no? We really need the rest rewrite. 🫣

mmschlk (changes requested):
I am still scared about the log weights, but I think you hit all consumers and it should be fine. So dope that we have numerical stability now!

I spotted one major issue: the fallback case in the try except in the selector of the tree proxies is not hit. If you ask for XGBoost, do not have it but have lightgbm, it will not get selected but will go straight to the tree case.

Further, I marked a couple of points worth refactoring before we can merge this. Despite this, I did not spot any serious issues.

CHANGELOG.md
Advueu963: They were wired actually also for the AgnosticExplainer^^

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
 # Changelog
 
+## v1.5.2 (unknown)
```
mmschlk:
```suggestion
## v1.5.2 (2026-06-12)
```

src/shapiq/approximator/proxy/_routes.py
```diff
@@ -0,0 +1,282 @@
...
+
+if TYPE_CHECKING:
+    from sklearn.linear_model._base import LinearModel
+
+    from shapiq.typing import CoalitionMatrix, GameValues
+
+ValidProxySHAPIndices = Literal["k-SII", "FSII", "FBII", "SII", "SV", "BV"]
+
+
+def _base_estimator(proxy_model: object) -> object:
```
mmschlk: a bit weird with the object to object mapping.

src/shapiq/approximator/proxy/_models.py
```diff
@@ -0,0 +1,421 @@
...
+    """Select a proxy model based on a string identifier.
+
+    Recognized identifiers are the members of :data:`ProxyLiteral`. The gradient-boosting tags
+    (``"xgboost"``, ``"lightgbm"``) try the requested backend, then the other boosting backend,
+    and finally fall back to a :class:`~sklearn.tree.DecisionTreeRegressor` when neither is
+    installed. Whenever a fallback happens, the accumulated reasons are reported in a single
+    warning rather than one warning per step.
+    """
+    # Reasons are accumulated so that any fallback path emits exactly one combined warning.
+    reasons: list[str] = []
```
mmschlk, line 378: Would be nice to print the list of available literals in the warnings as well (aka. ProxyLiteral).

mmschlk (changes requested):
The issues I had last time are fixed! :) I marked two things.

src/shapiq/approximator/proxy/_models.py
```diff
@@ -416,6 +416,8 @@ def _select_base_proxy_via_string(proxy_str: ProxyLiteral, random_state: int | N
         return LinearRegression()
     else:
         reasons.append(f"Proxy model '{proxy_str}' is not recognized.")
-
+    reasons.append(
+        "Currently available proxy models are 'xgboost', 'lightgbm', 'tree', and 'linear'."
```
mmschlk: I do not like that these strings are now hardcoded here. They should be inferred from the `ProxyLiteral` defined above.

src/shapiq/approximator/proxy/_routes.py
```diff
@@ -34,13 +34,13 @@
...
 
-def _base_estimator(proxy_model: object) -> object:
+def _base_estimator(proxy_model: object) -> ProxyModel:
     """Return a proxy's *base* estimator: an HPO wrapper's ``.estimator``, else the model itself.
 
     Used by :func:`fit_proxy` to pick the feature transform from the base estimator's type before
     the (possibly wrapping) model is fit.
     """
-    return getattr(proxy_model, "estimator", proxy_model)
+    return cast("ProxyModel", getattr(proxy_model, "estimator", proxy_model))
```
mmschlk: This seems like a bandaid fix. Is there no way of "getting" the `estimator` attribute from the proxy_model input? for this the input type would need to be changed and presumably the underlying HPOmodel / proxy model classes?

mmschlk (changes requested):
Have still two questions there.

mmschlk (approved):
Nice!

PR #548, opened 2026-06-12
Add ShaplEIG: Bayesian experimental design approximator

david-rundel:
ShaplEIG fits a Gaussian process surrogate with a weighted Hamming product kernel on queried coalition values and sequentially selects coalitions by maximizing the closed-form expected information gain about the Shapley values. The Shapley structure is handled via elementary-symmetric-polynomial identities of the kernel's generating polynomials, so the 2^n coalition space is never enumerated.

- new optional extra 'shapleig' (torch/gpytorch/botorch/linear_operator) with lazy import and an informative placeholder when not installed
- shapiq/approximator/shapleig/: approximator + flat Hamming GP surrogate + self-contained ESP math core (controlled copy of the validated reference implementation, kept in sync)
- order-1 SV InteractionValues output (min_order=0, baseline = empty coalition value); SV-variance return left as a commented scaffold until InteractionValues supports uncertainty information
- tqdm progress bar over the BED iterations
- registered in SV_APPROXIMATORS and the public API; unit tests covering output format, determinism, and convergence to ExactComputer ground truth

## Motivation and Context

This PR adds **ShaplEIG**, an approximator for Shapley values based on Bayesian experimental design (BED), accompanying the paper *"ShaplEIG: Bayesian Experimental Design for Shapley Value Estimation"* (Rundel et al., 2026).

The scope was pre-agreed with @mmschlk and @Advueu963: a black-box, end-to-end approximator in `shapiq/approximator/shapleig/` behind a new optional dependency group `shapleig`.

  **Open questions for the maintainers:**

  - The `shapleig` extra is pinned to the exact versions used for the published experiments
  (`torch==2.9.1`, `gpytorch==1.14`, `botorch==0.14.0`, `linear_operator==0.6`). Is this ok?
  - The surrogate posterior also provides the marginal SV *variances*; the code contains a commented
  scaffold to expose them via an `approximate_with_variance` once `InteractionValues` supports
  uncertainty information (future work, as discussed).

---

## Public API Changes

-   [ ] No Public API changes

  - New public class `shapiq.ShaplEIG` (= `shapiq.approximator.shapleig.ShaplEIG`), registered in
  `SV_APPROXIMATORS`.
  - New optional dependency group `shapleig` in `pyproject.toml`.
  - No changes to any existing API. `import shapiq` works without the extra installed (the optional
  dependencies are imported in `ShaplEIG.__init__`, which raises an `ImportError` pointing to `pip
  install 'shapiq[shapleig]'` when they are missing).

---

## How Has This Been Tested?

  - New unit tests in `tests/shapiq/tests_unit/tests_approximators/test_approximator_shapleig.py`:
  output format (`InteractionValues`, order 1, `min_order=0`, baseline = empty-coalition value),
  determinism under `random_state`, convergence to `ExactComputer` ground truth with growing budget,
  budget/candidate-set validation, sampled-candidate-subset and warmstart configurations.
  - Verified **bit-identical** against the validated reference implementation of the paper on the
  same game/seed/protocol (default and warmstart configurations).
  - Full local unit suite passes (852 passed; the only failures are pre-existing `[spex]` tests
  caused by the `sparse` extra not being installed locally, identical on `main`).

---

## Checklist

-   [ ] Documentation has been updated (if the public API or usage changes). (Not actively. Does the API reference pick 
  `ShaplEIG` up via the `shapiq.approximator` autosummary?)

---

### DIFF
--- a/src/shapiq/approximator/shapleig/_surrogate.py
+++ b/src/shapiq/approximator/shapleig/_surrogate.py
@@ -75,23 +75,34 @@ class HammingGP:
         }
         self.amount_restarts = amount_restarts
 
-        # The model must be CONSTRUCTED under a float64 default dtype: torch
-        # creates parameters, prior/constraint buffers, and the constraints'
-        # initial values in the default dtype before botorch casts the model
-        # to the (float64) data. Under a float32 default these starting
-        # values are truncated and L-BFGS converges to (slightly) different
-        # hyperparameters, changing the coalition selections. The default is
-        # restored right after construction; all downstream computations are
-        # float64 because the data and parameters are.
-        previous_dtype = torch.get_default_dtype()
-        torch.set_default_dtype(torch.float64)
+        torch_default_dtype = torch.get_default_dtype()
         try:
+            # The model must be CONSTRUCTED under a float64 default dtype: torch
+            # creates parameters, prior/constraint buffers, and the constraints'
+            # initial values in the default dtype before botorch casts the model
+            # to the (float64) data. Under a float32 default these starting
+            # values are truncated and L-BFGS converges to (slightly) different
+            # hyperparameters, changing the coalition selections. The default is
+            # restored right after construction; all downstream computations are
+            # float64 because the data and parameters are
+            torch.set_default_dtype(torch.float64)
             self._build(train_X, train_Y)
         finally:
-            torch.set_default_dtype(previous_dtype)
+            torch.set_default_dtype(torch_default_dtype)
 
     def _build(self, train_X: torch.Tensor, train_Y: torch.Tensor) -> None:
         """Construct the botorch model (called under float64 default dtype)."""
+        if torch.get_default_dtype() is not torch.float64:
+            msg = (
+                "HammingGP._build must run under a float64 default dtype "
+                f"(found {torch.get_default_dtype()}); it is invoked from "
+                "__init__ inside a torch.set_default_dtype(torch.float64) "
+                "context. Constructing under float32 truncates the model's "
+                "parameter/prior/constraint initial values and silently "
+                "changes the fitted hyperparameters."
+            )
+            raise RuntimeError(msg)
+
         fixed_noise_level = self._init_kwargs["fixed_noise_level"]
         min_lengthscale = self._init_kwargs["min_lengthscale"]
         min_inferred_noise_level = self._init_kwargs["min_inferred_noise_level"]
--- a/src/shapiq/approximator/shapleig/_surrogate.py
+++ b/src/shapiq/approximator/shapleig/_surrogate.py
@@ -75,6 +75,27 @@ class HammingGP:
         }
         self.amount_restarts = amount_restarts
 
+        # The model must be CONSTRUCTED under a float64 default dtype: torch
+        # creates parameters, prior/constraint buffers, and the constraints'
+        # initial values in the default dtype before botorch casts the model
+        # to the (float64) data. Under a float32 default these starting
+        # values are truncated and L-BFGS converges to (slightly) different
+        # hyperparameters, changing the coalition selections. The default is
+        # restored right after construction; all downstream computations are
+        # float64 because the data and parameters are.
+        previous_dtype = torch.get_default_dtype()
+        torch.set_default_dtype(torch.float64)
+        try:
+            self._build(train_X, train_Y)
+        finally:
+            torch.set_default_dtype(previous_dtype)
+
+    def _build(self, train_X: torch.Tensor, train_Y: torch.Tensor) -> None:
+        """Construct the botorch model (called under float64 default dtype)."""
+        fixed_noise_level = self._init_kwargs["fixed_noise_level"]
+        min_lengthscale = self._init_kwargs["min_lengthscale"]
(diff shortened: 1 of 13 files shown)


src/shapiq/approximator/shapleig/_surrogate.py
```diff
@@ -0,0 +1,187 @@
...
+        # values are truncated and L-BFGS converges to (slightly) different
+        # hyperparameters, changing the coalition selections. The default is
+        # restored right after construction; all downstream computations are
+        # float64 because the data and parameters are.
+        previous_dtype = torch.get_default_dtype()
+        torch.set_default_dtype(torch.float64)
+        try:
+            self._build(train_X, train_Y)
+        finally:
+            torch.set_default_dtype(previous_dtype)
```
mmschlk:
```suggestion
        torch_default_dtype = torch.get_default_dtype()
        try:
            # The model must be CONSTRUCTED under a float64 default dtype: torch
            # creates parameters, prior/constraint buffers, and the constraints'
            # initial values in the default dtype before botorch casts the model
            # to the (float64) data. Under a float32 default these starting
            # values are truncated and L-BFGS converges to (slightly) different
            # hyperparameters, changing the coalition selections. The default is
            # restored right after construction; all downstream computations are
            # float64 because the data and parameters are
            torch.set_default_dtype(torch.float64)
            self._build(train_X, train_Y)
        finally:
            torch.set_default_dtype(torch_default_dtype)
```

I also do not know a more beautiful way of handling this then at this point without setting the default then here...

src/shapiq/approximator/shapleig/_surrogate.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def _build(self, train_X: torch.Tensor, train_Y: torch.Tensor) -> None:
+        """Construct the botorch model (called under float64 default dtype)."""
```
mmschlk, line 94: If this step only can run with _build under a float64 default in torch, maybe we should then have a ValueError / RuntimeError be thrown if it is run with a wrong torch default value specified. This would make the downstream use a bit more explicit than just having a code comment.

Something like:

if default wrong:
   raise ValueError(wrong default)

This would ensure some code changes later to not break this silently.

src/shapiq/approximator/shapleig/shapleig.py
```diff
@@ -0,0 +1,446 @@
...
+        >>> from shapiq.approximator import ShaplEIG
+        >>> game = DummyGame(n=7, interaction=(1, 2))
+        >>> approximator = ShaplEIG(n=7, random_state=42, show_progress=False)
+        >>> values = approximator.approximate(budget=50, game=game)
+        >>> values.index, values.max_order
+        ('SV', 1)
+
+    References:
+        Rundel, D., et al. (2026). ShaplEIG: Bayesian Experimental Design for
+        Shapley Value Estimation.
```
mmschlk: We also use this bibliography environment actually in the docs. You could add this reference there and then refer to it here without specifying the Reference as new text. Then we have only one location where we need to maintain it. Did you see some examples in the code doing it like you did here? I thaught, I wired the references in correctly now.

https://github.com/mmschlk/shapiq/blob/main/docs/source/references.bib

src/shapiq/approximator/shapleig/shapleig.py
```diff
@@ -0,0 +1,446 @@
...
+        no_refit_step: bool,
+        prev_selected_idx: int | None,
+    ) -> torch.Tensor:
+        """EIG utilities of the candidates under the current surrogate.
+
+        In a no-refit step the hyperparameters are unchanged: the cached
+        A·K(Z,Z)·Aᵀ is reused, the column of the last queried coalition is
+        appended to A·K(Z,X), and the selected candidate's column is dropped
+        from A·K(Z,W).
+        """
```
mmschlk:
```suggestion
         r"""EIG utilities of the candidates under the current surrogate.

        In a no-refit step the hyperparameters are unchanged: the cached
        :math:`A K(Z, Z) A^\top` is reused, the column of the last queried
        coalition is appended to :math:`A K(Z, X)`, and the selected
        candidate's column is dropped from :math:`A K(Z, W)`.
        """
```

The unicode math stuff will not render correctly in the docstrings. The "r" makes it a raw string wich you need for latex. And then the :math: keyword in ``.

I did not go through the remaining code base. Maybe you could double check this again weather some additional changes would be required?

mmschlk (changes requested):
Nice work! I only marked a couple of small nitpicks.

One additional thing; could you create a string option for ShaplEIG (maybe "eig") inside the explainers `src/shapiq/explainer/configuration.py` and [`ValidApproximatorTypes`]( https://github.com/mmschlk/shapiq/blob/1b8b64339e9e4238ec79c1cdd349f967b4b0301a/src/shapiq/explainer/configuration.py#L29-L30). You do not have to wire it into the auto logic, but then users through the explainer api can also choose it directly. :)

mmschlk (approved):
As discussed, this is a wonderful contribution and in my perspective ready to go. :)

Advueu963 (approved):
Amazing work! Reall cool to have it now in shapiq

PR #550, opened 2026-06-16
GraphShapIQ Implementation

Proman9000:
## Motivation and Context

GraphSHAP-IQ is an algorithm for computing exact Shapley interaction values for Graph Neural Networks (GNNs), by Muschalik et. al https://arxiv.org/abs/2501.16944. It is the graph analogue of TreeSHAP-IQ and exploits the locality of message passing in GNNs. A node's embedding after L layers depends only on its L-hop neighborhood (its receptive field), which means Moebius interactions between nodes are zero unless all nodes fall within a single receptive field. This reduces the complexity from O(2^n) to O(n * 2^r), where r is the maximum receptive field size.

The algorithm works by:
1. Computing the L-hop receptive field for each node via BFS
2. Enumerating only coalitions within these receptive fields (not all 2^n)
3. Batch-evaluating these coalitions through the GNN game
4. Computing Moebius coefficients from the evaluations
5. Converting Moebius coefficients to the target interaction index via shapiq's existing `MoebiusConverter`

This PR adds a new `shapiq.graph` subpackage implementing this algorithm.

- **`GraphGame`**: This class wraps a GNN model (GCN, GIN, or GAT) together with an input graph. Players are graph nodes, and `value_function(coalitions)` evaluates the GNN's prediction for each coalition by masking absent nodes' features to a baseline while keeping the graph structure intact. Supports both classification and regression GNNs.
- **`GraphSHAPIQ`**: The core algorithm. Computes each node's L-hop receptive field via BFS, enumerates and deduplicates the union of powersets over these receptive fields, batch-evaluates them through a `GraphGame`, computes Moebius coefficients, and converts them to the target interaction index using the existing `MoebiusConverter`.
- **`GraphExplainer`**: This class integrates `GraphSHAPIQ` into shapiq's explainer framework, following the `TreeExplainer` pattern. Constructs the `GraphGame` internally and exposes a clean `explain(x)` API.

## Public API Changes

- [x] Yes, Public API changes (Details below)

New `shapiq.graph` subpackage exposing `GraphGame`, `GraphSHAPIQ`, and `GraphExplainer`.

## How Has This Been Tested?

The `GraphSHAPIQ` algorithm has been thoroughly tested, including correctness against `ExactComputer` on small graphs, the efficiency axiom, all three GNN architectures (GCN, GIN, GAT), and edge cases (disconnected graphs, single-node graphs, fully connected graphs). `GraphGame` has also been thoroughly tested. `GraphExplainer` and the L-Shapley algorithm have been only slightly tested so far. Overall test coverage is above 90%.

## Checklist

- [x] The changes have been tested locally.
- [ ] Documentation has been updated (if the public API or usage changes).
- [ ] An entry has been added to `CHANGELOG.md` (if relevant for users).
- [ ] The code follows the project's style guidelines.
- [x] I have considered the impact of these changes on the public API.


mmschlk:
Hello, thank you for your draft PR already! Just ping me here as soon as you want a PR review. :)

Note however, that for a review, the CI pipeline should really be green first (all tests pass, the code-quality checks are okay, and the docs building pipeline compiles). Otherwise it's hard for me to give you good feedback. :)

I can also approve the workflows from time to time if you ping me.

amndzdzdz:
@mmschlk could you approve the workflow please? The CI pipes should be green now

luffdavid:
> Hello, thank you for your draft PR already! Just ping me here as soon as you want a PR review. :)
> 
> Note however, that for a review, the CI pipeline should really be green first (all tests pass, the code-quality checks are okay, and the docs building pipeline compiles). Otherwise it's hard for me to give you good feedback. :)
> 
> I can also approve the workflows from time to time if you ping me.

@mmschlk the PR is now ready for review :)

src/shapiq/graph/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Docstring."""
```
mmschlk: write a real doctoring for the graph package

src/shapiq/graph/base.py
```diff
@@ -0,0 +1,167 @@
...
+import torch
+
+if TYPE_CHECKING:
+    from torch_geometric.data import Data
+
+from shapiq.game import Game
+
+
+class GraphGame(Game):
+    """A GraphSHAP-IQ explanation game for graph networks."""
```
mmschlk: very sparse documentation. Please explain this further. Each attribute needs to be documented. Follow the Google Style Guidelines for writing docstrings.

src/shapiq/graph/explainer.py
```diff
@@ -0,0 +1,316 @@
...
+    It is based on the GraphSHAPIQ algorithm by `Muschalik et al. (2025)` [Mus25]_, which efficiently
+    calculates any order Shapley Interactions for GNN-based predictions, by utilizing the unique
+    characteristics of GNNs, where only nodes in node i's l-hop neighborhood are considered for coalitions,
+    thus drastically reducing the total amount of coalitions to evaluate.
+
+    The GraphExplainer can be used with a variety of graph-based models, including
+    GCNs, GATs and GINS.
+
+    References:
+        .. [Mus25] Maximilian Muschalik and Fabian Fumagalli and Paolo Frazzetto and Janine Strotherm and Luca Hermes and Alessandro Sperduti and Eyke Hüllermeier and Barbara Hammer. Exact Computation of Any-Order Shapley Interactions for Graph Neural Networks. https://arxiv.org/abs/2501.16944
```
mmschlk: we have this reference.bib file in the docs. Please add the reference there and cite it in the docstrings similar to the other classes in shapiq.

pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
     "networkx",
     "colour",
     "pillow",
+    "torch-geometric>=2.8.0",
```
mmschlk: this needs to be an optional dependency living in shapiq[graph]. check out the other alternative extra groups we have to bring your new dependencies into new groups

for more info seach for optional dependencies or extra dependencies

src/shapiq/graph/explainer.py
```diff
@@ -0,0 +1,316 @@
+"""The GraphExplainer class for the shapiq package."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Literal, cast, override
+
+import joblib
+import numpy as np
+from torch_geometric.data import Data
```
mmschlk: we need to move this import into an optional import where shapiq can also work without torch_geometric needing to be installed. Check out how we did it for example with the ShaplEIG approximator and the torch dependencies:

https://github.com/mmschlk/shapiq/tree/main/src/shapiq/approximator/shapleig

We need to throw an error if the Explainer cannot work and so forth ...

See also the comment with the optional dependencies in the pyproject.toml

src/shapiq/graph/base.py
```diff
@@ -0,0 +1,167 @@
...
+        class_index: int | None = None,
+        baseline_strategy: str | float | None = None,
+        normalize: bool = True,
+        verbose: bool = False,
+    ) -> None:
+        """Initialize the GraphGame."""
+        self._normalize = normalize
+        self.x_graph = x_graph.clone()
+
+        # Validierungen
```
mmschlk: please do not have any German comments. English only.

src/shapiq/graph/base.py
```diff
@@ -0,0 +1,167 @@
...
+
+class GraphGame(Game):
+    """A GraphSHAP-IQ explanation game for graph networks."""
+
+    def __init__(
+        self,
+        model: torch.nn.Module,
+        x_graph: Data,
+        *,
+        task: Literal["classification", "regression"] = "classification",
```
mmschlk: I do not necessarily see why it is important to know at this point weather its classification or regression. All this information can be carried inside the class_index, no? If it is None then regression (or default class 1) and if its specified then you can do whatever.

src/shapiq/graph/base.py
mmschlk: This is everywhere in the code. I just mark it here for brevity.

src/shapiq/graph/base.py
```diff
@@ -0,0 +1,167 @@
...
+                verbose=verbose,
+            )
+        else:
+            super().__init__(n_players=self.n_players, normalize=normalize, verbose=verbose)
+
+        # Aktualisiere normalization_value in der Instanz (falls nötig)
+        if normalize:
+            self.normalization_value = normalization_value
+
+    def _calculate_baseline(self, strategy: str | float | None) -> torch.Tensor:
```
mmschlk: Why is `strategy` a str or a float or None. I would expect choice parameters to always be strings, and more specifically literal choices (e.g. Literal["Option A"] ...).

src/shapiq/graph/base.py
```diff
@@ -0,0 +1,167 @@
...
+        """Berechnet die Baseline für Masking basierend auf der Strategie."""
+        if strategy is None:
+            warnings.warn(
+                "Baseline is not provided, baseline will be initialized as zero...", stacklevel=2
+            )
+            strategy = "zeros"
+
+        x = self.x_graph.x
+        if isinstance(strategy, torch.Tensor):
```
mmschlk: "strategy: str | float | None" from the types this should be unreachable ...

src/shapiq/graph/graphshapiq.py
```diff
@@ -0,0 +1,419 @@
...
+        incomplete_neighborhoods_lookup: dict[tuple[int, ...], int],
+        max_subset_size: int,
+        grand_coalition_prediction_node: NDArray[np.floating],
+    ) -> InteractionValues:
+        """Adjust Möbius coefficients to ensure efficiency.
+
+        The efficiency axiom states that the sum of all Möbius coefficients must equal
+        the grand coalition prediction:
+
+            sum_{S ⊆ N} m(S) = v(N)
```
mmschlk: I am placing this here as it's the first time I notice it, but for the docs to be rendered on all machines correctly, we cannot have math notation written like this with Unicode chars. For this, you need to use special math escape arguments such as in the rest of shapiq (for example [here](https://github.com/mmschlk/shapiq/blob/7e7ade2d47c9389475810f269ee1b3f657a4206c/src/shapiq/approximator/shapleig/shapleig.py#L380)). The keyword :math:`<your math notation>`

mmschlk: This is throughout the implementation and need to be consistent.

tests/shapiq/graph/test_graphshapiq.py
```diff
@@ -0,0 +1,980 @@
...
+        assert interactions.index == "k-SII"
+
+    def test_explain_with_gat_model(self, gat_graphshapiq):
+        """Test that explain works correctly with a GAT model."""
+        moebius, interactions = gat_graphshapiq.explain()
+        assert isinstance(moebius, InteractionValues)
+        assert isinstance(interactions, InteractionValues)
+        assert interactions.index == "k-SII"
+
+    def test_explain_matches_exact_computer_k_sii(self, gcn_graphshapiq, gcn_graph_game):
```
mmschlk, line 784: Really nice you have this test! As discussed the ExactComputer here is really the right way to do this..

However, I think there is an issue here: All of these comparisons (k-SII/SII/STII and the order-1 test, lines 782–832) run on the 4-node cycle where every node's 2-hop neighborhood already covers all 4 nodes. That's the *complete* case, where GraphSHAP-IQ degenerates to a plain full Möbius transform. The receptive-field truncation that is the whole point of graphshapiq and is never actually exercised. Please add an exact-match test on a graph where max_size_neighbors < n_players. 

The way I would do this is to actually sample a bunch of small graphs from a small dataset or (have some synthetic function to generate them) and just brute force a couple inside the test (as long as this is not too computationally heavy.

src/shapiq/explainer/utils.py
```diff
@@ -24,7 +24,7 @@
 )
 
 ExplainerTypes = Literal[
-    "tabular", "tree", "tabpfn", "game", "product_kernel", "knn", "wknn", "tnn"
+    "tabular", "tree", "tabpfn", "game", "product_kernel", "knn", "wknn", "tnn", "graph"
```
mmschlk, line 27: The "graph" entry you added to `get_explainers()` (utils.py:54) and ExplainerTypes is currently unreachable: `get_predict_function_and_model_type` has no GNN detection, so `shapiq.Explainer(gnn)` classifies the model as "tabular" and routes to TabularExplainer. Either add a guarded detection branch (e.g. isinstance(model, torch_geometric.nn.MessagePassing), or something similar. I am not sure if these GNNs can be easily "detected" like that) that returns "graph", or if you are not able to do have a good check / validation, then we should only allow GraphExplainer to be used directly. Then we need to drop it from `get_explainers()` and document that. Right now it's half-wired.

src/shapiq/graph/graphshapiq.py
```diff
@@ -0,0 +1,419 @@
...
+            additional_values[i] = gap
+            additional_lookup[neighborhood] = i
+
+        # Adjust the largest neighborhood to maintain efficiency
+        if additional_lookup:
+            additional_values[-1] = (
+                grand_coalition_prediction_node
+                - np.sum(moebius_coefficients.values)
+                - np.sum(additional_values[:-1])
+            )
```
mmschlk: I think there's a latent bug here. This overwrite forces the *last* incomplete neighborhood's coefficient to the global efficiency residual (v(N) − \sum möbius − \sum other gaps), which is only valid if the largest incomplete neighborhood actually **_IS_** the grand coalition N. On a disconnected graph the largest neighborhood is just one component, so the gap you correctly computed a few lines up gets clobbered by a global correction that doesn't belong to it. The total sum still satisfies efficiency, which I think is why the current tests don't catch it (see also the comment in the exact computer tests), but the individual per-interaction values are wrong whenever efficiency_routine=True on an incomplete + disconnected graph. Please guard the overwrite so it only applies when the last neighborhood spans all N players, but maybe this is also not so good to do at all, since this would only be true for graphs we cannot compute anything quicker with GraphSHAP-IQ anyways... Is this part of our original implementation as well?

src/shapiq/graph/l_shapley.py
```diff
@@ -0,0 +1,233 @@
...
+from scipy.special import binom
+
+from shapiq.interaction_values import InteractionValues
+from shapiq.utils import powerset
+
+if TYPE_CHECKING:
+    from .base import GraphGame
+
+
+class LShapley:
```
mmschlk: **Okay this is the big one**: I know why you got this function as well. However, we never really talked about LShapley and that had a reason. We do not really need or want it here. LShapley only works approximately and only for Shapley Values, no other interaction indices or anything. So LShapley would actually be a special approximator and not necessarily a Graph-only method. This is why I would actually not want LShapley to be part of the PR and remove it entirely from here. Feel absolutely free to use it inside your experiment and/or illustrations, but let's keep it completly out of the PR. This means you can delete this file here, the tests concerning it and most importantly removing it from the GraphExplainer definitions and usage.

mmschlk (changes requested):
Thank you very much for your PR and on your work regarding GraphSHAP-IQ and its implementation inside shapiq. From the core implementation, I think that this should be actually in a very good stage. I did not really find some core issues which would be breaking GraphSHAP-IQ's behavior dramatically (one smaller comment on the efficiency thingy). This is really good!

However, I marked a couple of points in some key files as comments that need to be addressed such that we can merge this PR. I also would like to name the important / general findings here as well:

1. LShapley should actually be removed as it matches poorly with the remaining stuff from GraphSHAP-IQ and should be done separately in a different manner.

2. Core dependencies change: We are not allowed to have torch_geometric as a new dependency and should handle this lazyily. Please take a look at we did this in other instances inside shapiq (for example ShaplEIG).

3. The documentation is rather sparse. Please extend the docstrings, and add some examples to the examples folder that will then be picked up inside shapiq's documentation (https://github.com/mmschlk/shapiq/tree/main/examples). Note, the examples should be rather quick and intutitve they should not need to run more than a couple of seconds. Otherwise for each PR and Test the CI needs to run considerably longer. Additionally, if I build the docs locally I cannot find your graph package in the docs: This is because it is not discovered in the api_reference (https://github.com/mmschlk/shapiq/blob/main/docs/source/api_reference.rst). You can build the docs locally: `uv run --no-sync sphinx-build --keep-going -b html docs/source docs/build/html`. For more info see the GitHub workflow building it: https://github.com/mmschlk/shapiq/blob/main/.github/workflows/ci.yml

src/shapiq/graph/__init__.py
zahntheo: Done.

src/shapiq/graph/base.py
zahntheo: Done.

src/shapiq/graph/explainer.py
zahntheo: Done.

src/shapiq/graph/base.py
zahntheo: Done.

pyproject.toml
zahntheo: Done.

src/shapiq/graph/base.py
zahntheo: Done.

src/shapiq/graph/graphshapiq.py
zahntheo: Done.

src/shapiq/graph/l_shapley.py
amndzdzdz: Done.

src/shapiq/graph/base.py
zahntheo: Done.

src/shapiq/graph/base.py
zahntheo: Done.

tests/shapiq/graph/test_graphshapiq.py
zahntheo: Done.

src/shapiq/explainer/utils.py
zahntheo: Done.

src/shapiq/graph/graphshapiq.py
zahntheo: Done.

zahntheo: It's in the paper: [https://github.com/FFmgll/GraphSHAP-IQ/blob/main/shapiq/explainer/graph/graphshapiq.py#L131 (line  ](https://github.com/FFmgll/GraphSHAP-IQ/blob/6756d9597971f3b27796017f7165224834636ec4/shapiq/explainer/graph/graphshapiq.py#L131)

src/shapiq/graph/explainer.py
zahntheo: Done.

mmschlk:
I see lot's of "done"s in the PR review but I do not see any changes? Did you already commit and push them onto the PR? Is it a good time for another review?

luffdavid:
> I see lot's of "done"s in the PR review but I do not see any changes? Did you already commit and push them onto the PR? Is it a good time for another review?

@mmschlk Sorry for the confusion! We marked the comments as "done" internally so that two people wouldn't end up working on the same comment at the same time. We've now gone through all of the feedback, addressed the comments, and pushed the changes to the PR. Also, it includes now the C++ Part from @MaxFelix1008 .

Could you check again the PR? Thanks! 🙂

amndzdzdz:
@mmschlk could you review the changes please? :)

mmschlk:
@amndzdzdz, I will review/make changes to the methodological aspects of this PR again after the 17th for preparing the final merge into shapiq. :)

mmschlk (changes requested):
The graph package is still not getting displayed in the existing documentation. For this you have to change the api reference list for sphinx. Also, there are still no example scripts inside the examples folder.

mmschlk:
I will be bringing this in for 1.8.0. The next release contains a lot of work on the shapiq.tree side and I do not want to mix concerns.

PR #557, opened 2026-06-28
Implement TextImputer with Flexible Players, Perturbations and Target Callables

samoger:
## Motivation and Context

This PR implements a generalized TextImputer that integrates naturally into the existing shapiq imputation pipeline and enables coalition-based explanations for a broad range of NLP models through a modular architecture.

The implementation is built around three independent components:
* Player strategies — define the granularity of explanation (subwords, words, named entities, syntactic chunks, and sentences).
* Perturbation strategies — specify how missing players are represented, including [MASK], [PAD], removal, neutral replacement, WordNet-based replacement, MLM infilling, and attention masking.
* Target callables — provide model-specific scoring interfaces for encoder classifiers, causal language models, and encoder-decoder (Seq2Seq) models.

In addition, the implementation includes batched coalition evaluation for efficient model inference and integrates directly with HuggingFace transformers.

Compared with a model-specific implementation, this design cleanly separates player definitions, perturbation methods, and model architectures, making the TextImputer extensible to future NLP models and perturbation strategies while remaining compatible with the existing shapiq workflow.

---

## Public API Changes

-   [x] Yes, Public API changes (backwards compatible)
TextImputer follows the same interface as other shapiq imputers: it is initialized once with a model and an input text, then evaluated by passing coalition matrices directly (imputer(coalitions)).
### Constructor
```python
from shapiq.imputer.text_imputer import TextImputer
imputer = TextImputer(
    model=model,
    tokenizer=tokenizer,
    text=text,
    player_level="word",
    perturbation_type="mask",
    model_type="encoder_classifier",
)
```
##Supported player definitions
```python
player_level="subword"
player_level="word"
player_level="named_entity"
player_level="chunk"
player_level="sentence"
```
##Supported perturbation strategies
```python
perturbation_type="mask"
perturbation_type="pad"
perturbation_type="removal"
perturbation_type="neutral"
perturbation_type="wordnet_neutral"
perturbation_type="mlm_infilling"
perturbation_type="attention_mask"
```
##Supported model interfaces
```python
model_type="encoder_classifier"
model_type="causal_lm"
model_type="seq2seq"
```
---
##Optional configuration
```python
# Encoder classifiers
class_idx=1
output_type="logit"

# Causal LMs
target_label="good"
prompt_template="Review: {text}\n\nSentiment:"

# Seq2Seq
normalize_target_logprob=True

# MLM infilling
mlm_model_name="bert-base-uncased"
mlm_num_samples=100
```
### How Has This Been Tested?
Implemented three dedicated test suites covering the TextImputer components:
- `test_text_imputer.py`
  - Player strategies
  - Perturbation strategies
  - Target callables
  - TextImputer orchestration
  - Factory functions
  - Error handling
  
- `test_text_imputer_seq2seq.py`
  - Seq2Seq target callable
  - Target-sequence scoring
  - Teacher-forcing evaluation

- `test_text_imputer_attention.py`
  - Attention-mask perturbation
  - Attention-mask input construction
  - Encoder-classifier support
  - Causal-LM support
  - Seq2Seq support

---

## Checklist

---


src/shapiq/imputer/text_imputer.py
```diff
@@ -0,0 +1,1999 @@
...
+            batch_size=input_ids.shape[0],
+        )
+
+
+# =============================================================================
+# TEXT IMPUTER
+# =============================================================================
+
+
+class TextImputer:
```
mmschlk: This is the central architectural point. Every imputer in shapiq subclasses Imputer (which subclasses Game, see `MarginalImputer`/`BaselineImputer` in `imputer/base.py`). Yours is a bare `class TextImputer:` that imports nothing from shapiq, so it reimplements `value_function`/`n_players`/`__call__` by hand and inherits none of the framework: no `empty_coalition`/`grand_coalition`, no `normalization_value`, no precompute/save/load, and it can't be handed to TabularExplainer (which requires an `Imputer`). The whole point of an imputer is to plug into that machinery. Please subclass Imputer and implement the contract rather than paralleling it.

src/shapiq/imputer/text_imputer.py
```diff
@@ -0,0 +1,1999 @@
...
+        all_scores = []
+
+        for start in range(0, len(inputs), self.batch_size):
+            batch = inputs[start : start + self.batch_size]
+            batch_scores = self.target_callable.predict_from_inputs(batch)
+            all_scores.append(batch_scores)
+
+        return np.concatenate(all_scores)
+
+    def value_function(
```
mmschlk: value_function returns raw model scores with no empty_prediction and no normalization. Compare `MarginalImputer.value_function`, which sets `outputs[empty coalition] = self.empty_prediction`, and `Game.__call__,` which subtracts `normalization_value`. Shapley/SII approximators assume v(\emptyset)=0 (or a centered game); without it, your attributions won't sum to f(grand_coalition) − f(\emptyset) and the efficiency axiom fails. Right now an all-zero coalition just sends an all-[MASK]/empty string to the model and returns whatever it scores as the baseline. Compute the empty- and full-text scores in `__init__` and set `normalization_value = empty score`. This is a correctness issue for the shapiq computations down the line and not just an API nicety.

pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
     "networkx",
     "colour",
     "pillow",
+    "nltk>=3.9.4",
```
mmschlk: Two packaging problems here. (1) The module does top-level `import torch` (text_imputer.py:8) and `from transformers import ...` (line 13), but neither is an installable dependency, as they live only in the uv `all_ml`/dev group, which `pip install shapiq` never sees. So `from shapiq.imputer.text_imputer import TextImputer` crashes with a raw ModuleNotFoundError even for the mask/pad/removal perturbations that need no ML. (2) You added nltk to the CORE dependencies here, forcing it on every shapiq user. Our convention is an optional extra with a guarded import + friendly message — see approximator/shapleig (torch/gpytorch in [project.optional-dependencies].shapleig, wrapped in try/except with an "install the extra" error). Please add a `text` extra (transformers, torch, nltk) and guard the imports the same way.

tests/shapiq/tests_unit/tests_imputer/test_text_imputer.py
```diff
@@ -0,0 +1,886 @@
+"""Fast mock-based unit tests for shapiq TextImputer."""
```
mmschlk, line 1: Really nice job keeping the suite fully offline (I ran all 81 tests with HF_HUB_OFFLINE=1 and no NLTK data — 1.27s, all pass)! This is really really good! Very nice. Yet, I also spotted two gaps: (1) the top-level `from shapiq.imputer.text_imputer import ...` hard-imports transformers/torch, which aren't core deps, so on a plain install this is a *collection error*, not a skip. Add pytest.importorskip("transformers") / importorskip("torch") at the top of all three files, like tests_approximators/test_approximator_shapleig.py does. (2) Every model and tokenizer is a mock, so the tests encode your *assumptions* about the HF API (.logits shape, encoder outputs); they'd stay green even if the real code is wrong. Register a `slow` marker and add ONE opt-in end-to-end test against a tiny checkpoint (e.g. hf-internal-testing/tiny-random-*) to anchor the mocks to reality.

tests/shapiq/tests_unit/tests_imputer/test_text_imputer.py
```diff
@@ -0,0 +1,886 @@
...
+    ChunkPlayerStrategy,
+    EncoderClassifierCallable,
+    MaskTokenPerturbation,
+    MLMInfillingPerturbation,
+    NamedEntityPlayerStrategy,
+    NeutralPerturbation,
+    PadTokenPerturbation,
+    RemovalPerturbation,
+    SentencePlayerStrategy,
+    Seq2SeqCallable,
```
mmschlk: A few hygiene items: `uv run pre-commit run --all-files` (our 3s gate) currently fails: ruff F401, `Seq2SeqCallable` imported but unused here. Please run it before submitting. While you're in there: text_imputer.py:590 has a stray `"""Return empty string."""` sitting *after* the return in RemovalPerturbation.perturb (dead second docstring), and text_imputer.py:1733 says "Seq2seq is currently a placeholder" even though Seq2SeqCallable is fully implemented. Contradictory docs like that erode trust in the rest. Also worth adopting Google-style Args/Returns sections consistently (the 35-parameter __init__ documents none of its params).

src/shapiq/imputer/text_imputer.py
```diff
@@ -0,0 +1,1999 @@
+"""Generalized text imputer for coalition-based text explanations."""
```
mmschlk: One file holds ~7 player strategies, ~7 perturbation strategies, 4 model callables, 3 factories, and the imputer. Every other imputer is its own file, and heavier subsystems are subpackages (imputer/vision/, approximator/shapleig/). Please make this imputer/text/ with possibly (or others), players.py / perturbations.py / callables.py / imputer.py. This also lets the nltk-only and torch-only imports live in separate modules so a missing dep only disables the relevant path. Relatedly, TextImputer isn't in imputer/__init__.py's __all__, so it isn't a first-class imputer; the tests can only reach it via the private path.

src/shapiq/imputer/text_imputer.py
```diff
@@ -0,0 +1,1999 @@
...
+    A coalition uses ``1`` for a kept player and ``0`` for a missing player.
+    Missing players are replaced by the supplied perturbation strategy.
+    """
+
+    @abstractmethod
+    def get_players(self) -> list[str]:
+        """Return player list."""
+
+    @abstractmethod
+    def coalition_to_text(
```
mmschlk: The concrete coalition_to_text implementations in SubwordPlayerStrategy (120), WordPlayerStrategy (178), NamedEntityPlayerStrategy (257), and ChunkPlayerStrategy (338) are near-identical: same `len(coalition) == n_players` guard, the same enumerate → keep-or-perturb loop that builds an output list and passes context={"players", "coalition", "mask_index": idx} to the perturbation strategy, and the same `if replacement != ""` append. The ONLY thing that differs is the final join — SubwordPlayerStrategy uses tokenizer.convert_tokens_to_string(output), the others use " ".join(output).

That's a textbook template-method: put the shared loop here in BasePlayerStrategy.coalition_to_text (make it concrete instead of abstract) and have subclasses override just a small `_join(parts: list[str]) -> str` hook. SentencePlayerStrategy passes context=None, so add a class-level flag (e.g. `_passes_context`) for that one case. Right now the same ~25-line body is copied four times, which means any fix has to be applied four times — including the detokenization bug I flagged separately: once the loop is centralized, that fix lives in a single WordPlayerStrategy._join override rather than being repeated across every word-level strategy.

src/shapiq/imputer/text_imputer.py
```diff
@@ -0,0 +1,1999 @@
...
+    """
+
+    def __init__(
+        self,
+        tokenizer: PreTrainedTokenizerBase,
+    ) -> None:
+        """Initialize attention-mask perturbation."""
+        self.tokenizer = tokenizer
+
+    def perturb(
```
mmschlk: This subclasses BasePerturbationStrategy but perturb() just returns the input unchanged. The real work is in a different, tensor-returning interface (evaluate/build_*). Yet it's registered in the same create_perturbation_strategy factory as the string-returning strategies, so anyone selecting "attention_mask" and routing it through the normal coalition_to_text path gets silently unperturbed text. So we have wo incompatible mechanisms behind one abstraction. A solution to this is to give the attention/tensor-based perturbations their own base class and registry.

mmschlk (changes requested):
Really nice submission. Particularly the tests are very nice: 80+ genuinely substantive tests running fully offline in ~1.3s. This is really a hard part of testing ML code. The strategy/factory design is clean and thoughtful too! Great job.

Yet, I'm marking this "request changes" as there are a couple of issues to work through (mostly about fitting into shapiq's Imputer framework), which I've left as inline comments. Really nice work overall.

src/shapiq/imputer/text_imputer.py
samoger: Thanks for pointing this out. TextImputer now subclasses Imputer and calls its base initializer, so it is integrated into the existing Imputer / Game framework. The custom n_players and __call__ implementations were removed in favor of the framework-provided n_features and Game.__call__. The implementation also now uses the inherited empty_coalition and grand_coalition and participates in the framework’s normalization and precompute/save/load machinery.

src/shapiq/imputer/text_imputer.py
samoger: Thanks for pointing this out. We reworked the value-function handling to follow the Game / Imputer normalization flow. During initialization, TextImputer now evaluates and stores the predictions of the inherited empty_coalition and grand_coalition as empty_prediction and full_prediction. The empty prediction is used as normalization_value, so calls through Game.__call__() center the game at the empty coalition. value_function() also explicitly maps empty coalitions to the cached empty_prediction, consistent with the other imputers. We separated the raw coalition evaluation into _evaluate_coalitions() so the reference predictions can be computed without depending on the normalized call path. As a result, the normalized game has v(emptyset) = 0 and the attribution target is f(grand_coalition) - f(emptyset).

pyproject.toml
samoger: Checked, we moved the text-specific dependencies out of the core dependencies and added a dedicated text optional extra containing torch, transformers, and nltk. We also added a shared _error.py module for the text package, which centralizes the optional-dependency import error and provides a clear installation hint via pip install 'shapiq[text]'. The individual text modules now guard their runtime optional imports and raise this shared friendly error when a required text dependency is unavailable. This keeps a plain shapiq installation independent of the text stack while providing an explicit installation path for TextImputer.

tests/shapiq/tests_unit/tests_imputer/test_text_imputer.py
samoger: Thanks, we addressed both testing gaps. The text test modules now call pytest.importorskip("torch") and pytest.importorskip("transformers") before importing TextImputer, so environments without the optional text dependencies skip these tests during collection instead of failing with a ModuleNotFoundError. We also added a registered slow marker and one opt-in end-to-end test using the real hf-internal-testing/tiny-random-bert tokenizer and sequence-classification model. This test constructs a real TextImputer and evaluates coalitions through the actual Hugging Face tokenizer/model interface, while the default unit-test suite remains fast and mock-based. The default run currently reports 42 passed and 1 skipped, and the opt-in slow test passes separately when enabled with RUN_SLOW_TESTS=1.

tests/shapiq/tests_unit/tests_imputer/test_text_imputer.py
samoger: Checked, we cleaned up the hygiene issues mentioned here: the unused Seq2SeqCallable test import was removed, the stray string and unreachable second docstring were deleted, and the outdated Seq2Seq placeholder description was updated to reflect the implemented Seq2SeqCallable. The related tests are passing after these changes. We will also run the full pre-commit suite once more before the final push.

src/shapiq/imputer/text_imputer.py
samoger: Done, we split the previous monolithic implementation into separate players.py, perturbations.py, callables.py, and imputer.py modules, with tensor-based perturbations separated as well. This also allows the optional text dependencies to be handled within the relevant text modules with a shared friendly import error. TextImputer is now exported through both shapiq.imputer.text and shapiq.imputer, so it can be imported as a first-class imputer via from shapiq.imputer import TextImputer.

src/shapiq/imputer/text_imputer.py
samoger: Done, we refactored the player strategies using the suggested template-method approach. The shared coalition validation, keep-or-perturb loop, context construction, and replacement handling now live in BasePlayerStrategy.coalition_to_text(). Individual player strategies only override the _join() hook for their reconstruction behavior. We also added the _passes_context class-level flag so SentencePlayerStrategy can opt out of perturbation context while the other strategies share the common implementation.

src/shapiq/imputer/text_imputer.py
samoger: Done, we separated the two perturbation mechanisms as suggested. AttentionMaskPerturbation no longer subclasses the string-based BasePerturbationStrategy; it now lives in a separate tensor_perturbation.py module and subclasses a dedicated BaseTensorPerturbationStrategy with an evaluate() interface that returns model-ready tensor inputs. Tensor perturbations also have their own registry and factory, separate from the string-returning perturbation strategies. TextImputer now explicitly distinguishes "text" and "tensor" perturbation modes and routes tensor perturbations through evaluate() / predict_from_inputs() instead of coalition_to_text(). We also added validation to prevent text and tensor strategies from being mixed, avoiding the previous silent unperturbed-text path for attention_mask.

samoger:
Hi, we have now addressed all of the review comments and pushed the corresponding changes to the PR. We also replied to each comment individually to briefly explain the changes we made.

The PR should now be ready for another review. Thank you.

samoger:
Hi, We have fixed the CI issue from the previous run, so the import errors should now be resolved and the CI tests can be run again. Thank you!

XWTX42:
Hi! The NLTK WordNet issue has been fixed. The CI tests should pass now, so could you please rerun them? Thanks!

samoger:
Hi, we have fixed the issues from the previous CI test. Could you please run the tests again? Thanks!

samoger:
Hi! We have fixed the formatting issue that caused the CI failure. It was due to a branch mix-up after our final submission. Could you please re-run the CI test
s? Thank you!

mmschlk:
Thank you again for your work on the TextImputer! Unfortunately, for the time being we cannot directly merge this PR onto shape's main branch as it incurs too many changes that are not aligned with shapiq's current development progress and status. We will in the long run come back to these changes and also try to incorporate some of them into shapiq-v2. As for now however, I have to close this here.

PR #560, opened 2026-07-03
Oddshap improvements

FFmgll:
## Motivation and Context

Improvements in OddSHAP:
(1) Minimum budget changed from `interaction_factor * n` to `interaction_factor` by relaxing the condition that all individuals have to be in the Fourier regression support. Instead, the surrogate tree will now also detect the most important individuals in the low-budget regime. After `interaction_factor *n`, the interaction detection begins, which is similarl to the one described in the paper.
(2) Paired subsets are merged into a single row, which halves the number of rows in the Fourier regression without changing the solution to the least squares problem, as described in the paper.

## Public API Changes

-   [x ] No Public API changes

---

## How Has This Been Tested?

Expanded the unit test, and tested some tabular benchmarks.

## Checklist

---

### DIFF
--- a/src/shapiq/approximator/regression/oddshap.py
+++ b/src/shapiq/approximator/regression/oddshap.py
@@ -11,6 +11,7 @@ from collections import defaultdict
 from typing import TYPE_CHECKING, Any
 
 import numpy as np
+from scipy import linalg as scipy_linalg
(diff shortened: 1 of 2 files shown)


mmschlk (approved):
Nice!

PR #562, opened 2026-07-04
Conformance tests (task 4) – New Shapley Value Approximators

FabianK-Dev:
## Motivation and Context
This PR contributes the benchmark suite requested in task 4 that runs all approximators for SV (and, where relevant, for SII / k-SII) on a shared set of games. All of our new approximators (PolyShap, OddShap and LeverageShap) use this same benchmark suite to avoid duplicate code. Thus, we decided to separate it from the other approximator PRs and create a new PR. The task description only asks for plots but if possible I'll add a table print after the benchmark that compares the results in a table as well so you can get an overview quickly of the results.

**Disclaimer:**
This branch also contains changes to approximators (like LeverageSHAP) that were merged into this branch. This is only to allow us to test the changes of the benchmark. We can either revert those approximator changes later or wait for the approximator PRs to merge first into main.

---

## Public API Changes

-   [ ] No Public API changes

---

## How Has This Been Tested?

---

## Checklist

-   [ ] The changes have been tested locally.

---


mmschlk:
Yes! This is really nice for evaluating weather the approximators behave appropriately. Yet, we do not really want to have this benchmark suite in the main branch of shapiq, see this as supplemental material which is fine to keep on your branch. I will close this PR here accordingly.

PR #572, opened 2026-08-11
Adds WoodELF-IQ to shapiq

ron-wettenstein:
## Motivation and Context

Use Woodelf to speedup shapiq's TreeExplainer.
On large datasets or on path dependent interaction values, Woodelf's computation is going to be used behinds the schemes.

The feature implement explain_X directly and uses Woodelf vectorized computation to compute the values on X rows in a single run. 
On smaller datasets, TreeSHAPIQ is still used. We run experiments to discover the cross over between Woodelf and TreeSHAPIQ. 
These experiments are summarized in the htmls below:

        Path Dependent experiment:
        https://ron-wettenstein.github.io/TreeBranchMarks/benchmarks/reports/woodelf_vs_shapiq_path_dependent_experiment.html

        Interventional experiment:
        https://ron-wettenstein.github.io/TreeBranchMarks/benchmarks/reports/woodelf_vs_shapiq_experiment

The current cross over is n*m >= 100 and n >= 100 where n is the length of X and m is the length of the background dataset

---

## Public API Changes

-   [ ] No Public API changes

TreeExplainer now support Banzhaf values
The output format of the explain_X function was changed.

---

## How Has This Been Tested?
Added 2 tests. Also run some sanity checks.

---

## Checklist

---


ron-wettenstein:
Important stuff:

1. Critical - the flow for classifiers is broken. You use class_index=1 by default and Woodelf (and the shap package) use class_index=0. We need to discuss multi-class support. Building the woodelf tree format in shapiq will solve this and 4 below, but we did not want to do that... => FIXED now!

2. The Woodelf baseline_value (the value of the empty set) is broken in path dependent. We need to compute it directly somehow. => We still need to fix (Max)

3. We need to add woodelf_explainer as a dependency => We still need to add (Max)

Other stuff:
4. I do not support already parsed models (the case where TreeModel class is the input). When these are given Woodelf will not be used
5. I changed the output of explain_X, this output is now not the same as the output of Expainer explain_X.

ron-wettenstein:
I will solve 1 by enabling class_index selection in woodelf during the model parsing.
Have a working code, will finish this and deploy a new version tomorrow

ron-wettenstein:
Do you support this??

XGBoost default multi-class and LightGBM parse fine but store scalar leaves with one tree per class (leaf_vector_shape=[1 1], 40 trees for 10 rounds × 4 classes), which is the tree-filtering case you skipped.

If you revisit it: treelite exposes a per-tree class_id array ([0 1 2 3 0 1 2 3 …]), and filtering on it matched shap to ~2e-08 for all four classes on both models. The catch is the one in the option — it would change what a multi-class booster explains by default, from the sum of all class margins to class 0.

Let discuss this tomorrow

ron-wettenstein:
[woodelf_vs_shapiq_experiment.html](https://github.com/user-attachments/files/30973539/woodelf_vs_shapiq_experiment.html)
[woodelf_vs_shapiq_path_dependent_experiment.html](https://github.com/user-attachments/files/30973541/woodelf_vs_shapiq_path_dependent_experiment.html)

The experiments we based our cross over of 100 on

ron-wettenstein:
More TODOs:
- Have the woodelf format supported in plotting and aggregation functions
- Make class_index=None works in the shapiq parsing on XGBoost (It currently keep the whole tree instead of picking the one of class 1)

mmschlk:
Okay, I sat on this today and made some changes:

## The obvious things:
1. I added the dependency groups and warnings/errors that tell users about `shapiq[tree]`.
2. I fixed the code-quality stuff such that ty and ruff are happy. Yay.
3. I made `baseline_value` the property that gets computed once based off of a static method from InterventionalTreeExplainer (rename still pending, #574) which is also used in there to not make them drift.
4. Made tests go green

## The cooler things:
5. figured out that `treelite` seems to have some issues with macOS and incorporated that into the error messages and points user to a work-around. I raised an [issue with `treelite`](https://github.com/dmlc/treelite/issues/678) and will continue to monitor this if this is the correct solution.
6. made woodelf run also for `k-SII` by making the aggregate function safe also for woodelf's `dict[tuple[int, ...], np.ndarray]` output type
7. added a `backend` option to `TreeExplainer` which let's users force the behavior of `TreeExplainer`, which is automatically set to `"auto"`

Would love to hear what you say to this @ron-wettenstein! ❤️

mmschlk:
@ron-wettenstein, the docs failures came from our different return types making me second-guess our decision of just forking here. After chat with Claudius, I came to the conclusion that we should just do the batching now in one good solution so I just changed the output type of explain_X from `list[InteractionValues]` to `Sequence[InteractionValues]` and introduced your output format as the `InteractionValuesBatch` via `class InteractionValuesBatch(Sequence[InteractionValues]): ...` that materializes lazily. `InteractionValuesBatch.values` carry the data structure and getitems are helpful for downstream consumers. :)

Hope this also solves the doc building issues haha. 🤞

mmschlk:
I do not care for the codecov target. 😇

ron-wettenstein:
@mmschlk the treelite break on MacOS quite new right?
Can you check on you MacOS what previous versions do work?
I can pin in woodelf dependencies an older working version of treelite and solve the problem

.github/workflows/ci.yml
```diff
@@ -151,8 +151,16 @@ jobs:
...
+            -   if: matrix.os != 'macos-latest'
+                name: Run unit tests
                 run: uv run pytest "tests/shapiq"
+            -   if: matrix.os == 'macos-latest'
+                name: Run unit tests (macOS)
+                # treelite (needed by the woodelf fast path of the TreeExplainer) does not find
+                # Homebrew's keg-only libomp on its own (https://github.com/dmlc/treelite/issues/678).
+                # The DYLD variable must be set inline: SIP strips DYLD_* from the environment
+                # when the step's protected /bin/bash starts, so `env:`/GITHUB_ENV cannot carry it.
+                run: DYLD_FALLBACK_LIBRARY_PATH="$(brew --prefix libomp)/lib" uv run pytest "tests/shapiq"
```
ron-wettenstein, line 163: Did you create all this due to treelite failure? 
Maybe woodelf can depend on older treelite version the does not fail and then this can be removed

src/shapiq/tree/interventional/explainer.py
```diff
@@ -216,11 +216,7 @@ def __init__(
             self.index = "CUSTOM"
             self.look_up_table = self._build_custom_weight_table()
         # Compute the interventional baseline directly from the validated trees.
-        per_sample_predictions = np.array(
-            [sum(t.predict_one(ref) for t in self.tree) for ref in self.reference_data],
-            dtype=np.float64,
-        )
-        self.baseline_value = float(per_sample_predictions.mean())
+        self.baseline_value = self.compute_empty_prediction(self.tree, self.reference_data)
```
ron-wettenstein: Maybe don't compute it in the __init__ as you do not always use it. You can compute it once on demand and save the results in self._baseline_value (and have baseline_value as property as we discussed)

src/shapiq/tree/interventional/explainer.py
```diff
@@ -236,6 +232,29 @@ def __init__(
         if self.bool_tree:
             self._preprocess_boolean_tree()
 
+    @staticmethod
+    def compute_empty_prediction(trees: list[TreeModel], reference_data: np.ndarray) -> float:
```
ron-wettenstein, line 248: This is basically a predict function. It does 'trees.predict(reference_data)'.
Don't you have a prediction function in the framework that already do all this? 
If not, maybe this should sit near the TreeModel class or in same utils module.

src/shapiq/tree/explainer.py
```diff
@@ -86,14 +110,23 @@ def __init__(
                 indices such as ``"k-SII"``. Defaults to ``0``.
 
             index: The type of interaction to be computed. It can be one of
-                ``["k-SII", "SII", "STII", "FSII", "BII", "SV"]``. All indices apart from ``"BII"``
-                will reduce to the ``"SV"`` (Shapley value) for order 1. Defaults to ``"SV"``.
+                ``["SV", "SII", "k-SII", "BV", "BII"]``. All indices apart from the Banzhaf
+                indices ``"BV"`` and ``"BII"`` will reduce to the ``"SV"`` (Shapley value) for
```
ron-wettenstein: "All indices apart from the Banzhaf indices ``"BV"`` and ``"BII"`` will reduce to the ``"SV"``" => "Both ``"SII"`` and ``"k-SII"`` will reduce to the ``"SV"``"

src/shapiq/tree/explainer.py
```diff
@@ -113,28 +146,100 @@ def __init__(
...
+
         self._treeshapiq_explainers: list[TreeSHAPIQ] = []
         self._lineartreeshap_explainers: list[LinearTreeSHAP] = []
         self._interventional_explainer: InterventionalTreeExplainer | None = None
+        self._explainers_initialized = False
+
+        self._baseline_value: float | None = None
+
+    @property
+    def baseline_value(self) -> float:
```
ron-wettenstein, line 199: Nice!

src/shapiq/tree/explainer.py
```diff
@@ -113,28 +146,100 @@ def __init__(
...
-                # ``index`` (the local parameter) is already narrowed to ``TreeSHAPIQIndices``;
-                # ``self.index`` is the broader ``ExplainerIndices`` and would not type-check.
+                index = self.index
+                if index not in ("SV", "SII", "k-SII"):
+                    msg = (
+                        f"index='{index}' with mode='pathdependent' is only supported via the "
+                        "optional woodelf dependency (`pip install shapiq[tree]`), which is "
+                        "unavailable or does not support this configuration."
+                    )
+                    raise ValueError(msg)
```
ron-wettenstein: Can we ever reach this exception? You already raised it in the backend="shapiq" case and both backend "woodelf" and "auto" use the woodelf backend and should not call this function. 

What will happen if you continue with this function with "BV" will the code crash on some other exception? or will it finish and initialize wrong explainers? 
If it crashes then this exception can be delete (I think) if wrong explainer will be created it make sense to keep it.

src/shapiq/tree/explainer.py
```diff
@@ -173,6 +273,276 @@ def _can_use_lineartreeshap(self) -> bool:
...
+                BanzhafValues,
+                GeneralBanzhafInteractionValues,
+                GeneralShapleyInteractionValues,
+                ShapleyValues,
+            )
+            from woodelf.core.trees.parse_models import load_decision_tree_ensemble_model
+            from woodelf.woodelf_sparse import hybrid_woodelf
+        except ImportError as error:
+            msg = (
+                "The Woodelf fast path requires the optional 'woodelf-explainer' package. "
```
ron-wettenstein: You have these kinds of messages like 5-6 times in this class. Maybe unit some of them and save as a constant?

src/shapiq/tree/explainer.py
```diff
@@ -173,6 +273,276 @@ def _can_use_lineartreeshap(self) -> bool:
...
+            # typically treelite's macOS wheel failing to load libomp: treelite imports lazily
+            # inside woodelf, so a missing OpenMP runtime surfaces here as an OSError.
+            msg = (
+                "woodelf is installed but its treelite backend failed to load. On macOS this "
+                "usually means the OpenMP runtime is missing: install it with "
+                "`brew install libomp` and set "
+                "`DYLD_FALLBACK_LIBRARY_PATH=$(brew --prefix libomp)/lib`. "
+                "See https://github.com/dmlc/treelite/issues/678 for details."
+            )
+            raise RuntimeError(msg) from error
```
ron-wettenstein, line 421: The try except can surround only 
"""
loaded_model = load_decision_tree_ensemble_model(
                self.model, range(X.shape[1]), class_index=class_index
            )
"""
The problem occur there

src/shapiq/tree/explainer.py
```diff
@@ -173,6 +273,276 @@ def _can_use_lineartreeshap(self) -> bool:
...
+        """
+        return InteractionValuesBatch(
+            woodelf_result,
+            n_instances=n_instances,
+            n_players=n_players,
+            index=self._index,
+            max_order=self._max_order,
+            min_order=self._min_order,
+            baseline_value=self.baseline_value,
+        )
```
ron-wettenstein, line 507: Very cool!

src/shapiq/tree/explainer.py
```diff
@@ -294,9 +665,68 @@ def explain_function(  # type: ignore[override]
...
+            if woodelf_result is not None:
+                return self._woodelf_result_to_batch(
+                    woodelf_result, n_players=n_players, n_instances=len(X)
+                )
+
+        shapiq_results = super().explain_X(
+            X, n_jobs=n_jobs, random_state=random_state, verbose=verbose, **kwargs
+        )
+        return self._woodelf_result_to_batch(
+            self._cast_shapiq_results_to_woodelf_format(list(shapiq_results)),
```
ron-wettenstein: Isn't it a better and clearer option that casting shapiq result to woodelf format and than casting it again to shapiq batch format? 
Maybe build a series of shapiq results directly?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Highlights of new Features
 
+- adds a Woodelf fast path to `shapiq.tree.TreeExplainer` (Nadel and Wettenstein, 2026; Wettenstein et al., 2026): larger inputs are computed by the vectorized Woodelf algorithms, `TreeExplainer` now supports the Banzhaf indices `"BV"` and `"BII"`, `TreeExplainer.explain_X` explains all instances in one vectorized run and returns the new `InteractionValuesBatch` — a lazy sequence of one `InteractionValues` per instance whose `.values` attribute exposes the raw vectorized arrays (`Explainer.explain_X` is now typed as returning a `Sequence[InteractionValues]`). Requires the new optional `tree` dependency group (`pip install shapiq[tree]` — `woodelf-explainer`); without it, the explainer falls back to the shapiq implementation with a `WoodelfNotAvailableWarning`. The new `backend` parameter (`"auto"`, `"woodelf"`, `"shapiq"`) overrides the automatic routing [#572](https://github.com/mmschlk/shapiq/pull/572)
```
ron-wettenstein: The first sentence it too long, break it so 3-4 sentences.

src/shapiq/interaction_values.py
```diff
@@ -1243,3 +1243,128 @@ def _update_interactions_for_index(
...
+        Args:
+            item: The instance position in the batch, or a slice of positions.
+
+        Returns:
+            The instance's :class:`InteractionValues`, or a list of them for a slice.
+        """
+        if isinstance(item, slice):
+            return [self._materialize(row) for row in range(*item.indices(self.n_instances))]
+        row = item + self.n_instances if item < 0 else item
+        if not 0 <= row < self.n_instances:
```
ron-wettenstein: row < 0 or row >= self.n_instances
feels clearer here

ron-wettenstein: Or at least have "not (0 <= row < self.n_instances) "

pyproject.toml
```diff
@@ -61,6 +61,10 @@ keywords = [
 ]
 
 [project.optional-dependencies]
+tree = [
+    # required by the Woodelf fast path of shapiq.tree.TreeExplainer
+    "woodelf-explainer>=0.4.6",  # imports as `woodelf`; pulls in treelite>=4.0
```
ron-wettenstein, line 66: Hopelly many new versions are coming soon. 0.4.6 will seems so long ago haha

ron-wettenstein (commented):
See some comments

mmschlk:
> @mmschlk the treelite break on MacOS quite new right? Can you check on you MacOS what previous versions do work? I can pin in woodelf dependencies an older working version of treelite and solve the problem

No earlier treelite versions properly supported apple silicon, so pinning to an earlier version does not fix it, but in the issue is already fixed in [this PR](https://github.com/dmlc/treelite/pull/681) which will bring a new 4.8.0 version this month. Once this is there pinning treelite to 4.8.0 would probably be the safest and easiest solution. I will create a tracking issue for removing the error messages and so pointing to this.

src/shapiq/tree/interventional/explainer.py
mmschlk: Agreed! Thanks! Added TreeModel.predict and predict_ensemble as a static function. Baseline value calls this now in a lazy way. :) Fixed in 0769bf86.

src/shapiq/tree/explainer.py
mmschlk: Yes with "auto" pathdepenendt order >= 3 or larger tree depths

mmschlk: and catboost

mmschlk: aaaah I see ... but this would then not be the best solution to raise I feel ... let me think about this a bit

src/shapiq/tree/explainer.py
mmschlk: jup :D done

src/shapiq/tree/interventional/explainer.py
mmschlk: Yes agreed, baseline value calls this now in a lazy way. Fixed in 0769bf86

src/shapiq/interaction_values.py
mmschlk: Done in ab0d2391. Did "not (0 <= row < self.n_instances)" as the "row < 0 or row >= self.n_instances" is not ruff conform.

CHANGELOG.md
mmschlk: Done in ab0d2391. But do not stress this too much for the release I will anyways iterate over the changeling one more time.

src/shapiq/tree/explainer.py
mmschlk: Jup! Done in ab0d2391.

src/shapiq/tree/explainer.py
mmschlk: Removed in 1132092c. Okay, I opted to delete this one and make one additional change to TreeSHAP-IQ which should have been there already earlier. I now have a validation in TreeSHAP-IQ that raises on wrong indices, so we do not have to be overly protective here and let TreeSHAP-IQ just error instead of having so many guards.

src/shapiq/tree/explainer.py
mmschlk: Rewritten.

.github/workflows/ci.yml
mmschlk: Yes, but again, I will delete this as soon as treelite 4.8.0 is out and woodelf depends on this.

src/shapiq/tree/explainer.py
mmschlk: Okay. I basically still do something quite similar to what it was before but made it read and behave way nicer imo. I really want to use `super()`'s `explain_X` as it comes with the joblib loop and other goodies. So, re-implementing is not nice. We still get as a return just the list of interaction values (we would have gotten this anyways since all the TreeSHAP-IQ / LinearTreeSHAP, InterventionalTreeExplainer code is not built to do something else without a big rewrite). But now we can put this easily in the classmethod as a seperate constructor of the `InteractionValuesBatch` class that constructs the vectorized representation from the sequence one. I think this is from a callsite presepctive much cleaner and better seperation of concerns.

mmschlk:
From my side, this PR is ready to go. :)

src/shapiq/tree/explainer.py
ron-wettenstein: Yes sounds awesome max, it's the right solution

ron-wettenstein:
From my side too! Good work on my comments @mmschlk :)

mmschlk (approved):
Looks good from my side. Will also wait what @Advueu963 says as I have put a lot of opinions already in this one.

Advueu963 (approved):
Amazing job guys!

PR #577, opened 2026-08-15
Updating TreeParsing to Support Categorical Features

Advueu963:
## Motivation and Context

This PR adds support for categorical features in XGBoost, HistGradientBoosting and LightGBM to shapiq.
CatBoost is not supported, as this would require substantial rewriting, i.e. the TreeModel would require a specific hash function of CatBoost.
This is achieved by unifying the handling of categorical features to CSR in the `TreeModel`.
The idea is using `goes_left`, which is the central function to decide whether the split routes to the left or right child of the node. 

Furthermore, we fix a small bug in the class_index parsing of XGBoost.

## Public API Changes

-   [X] No Public API changes

---

## How Has This Been Tested?

Local Tests have been added, which check original tree prediction vs converted tree prediction.

## Checklist


src/shapiq/tree/base.py
mmschlk: `decision_function` no longer has any callers — this PR migrated `predict_one`, the interventional explainer, and the benchmark all to `goes_left`. Worse, it is categorical-blind: on a categorical node it compares the value against the sanitized NaN threshold, which is always `False`, so it always routes right regardless of the category set — silently reproducing exactly the mis-parsing class of bug this PR fixes for anyone who still calls it.

Can we either remove it, or delegate it to `goes_left` so it stays correct?

---
_Generated by [Claude Code](https://claude.ai/code)_

src/shapiq/tree/treeshapiq.py
mmschlk: `self._strict_lt` is dead now: its only consumer (the strict-lt threshold comparison in `_prepare_variables_for_order`) was replaced by `self._tree.goes_left(node_id, feature_value)` in this PR, and grep shows this line is the sole remaining occurrence. The four-line routing-convention comment above it now documents a mechanism that no longer exists, which risks misleading future maintainers into "fixing" routing through it.

Please drop the attribute and the comment block.

---
_Generated by [Claude Code](https://claude.ai/code)_

mmschlk (commented):
I went through the PR and my Claude spotted these two things here.

src/shapiq/tree/base.py
Advueu963: Solved in [2485ba1](https://github.com/mmschlk/shapiq/pull/577/commits/2485ba19ecf0515686f568e2db8c3628e8109cd6).

src/shapiq/tree/treeshapiq.py
Advueu963: Solved in [2485ba1](https://github.com/mmschlk/shapiq/pull/577/commits/2485ba19ecf0515686f568e2db8c3628e8109cd6).

src/shapiq/tree/base.py
mmschlk: So, we do not need decision_function anymore? We let go-left do everything, right?

mmschlk (approved):
Nice!

src/shapiq/tree/base.py
Advueu963: Yes exactly. It was replaced with go_left.

Issue #574, opened 2026-08-14
Breaking: rename `InterventionalTreeExplainer` to `InterventionalTreeSHAPIQ`


mmschlk:
### Describe the workflow you want to enable

Rename the `InterventionalTreeExplainer` to `InterventionalTreeSHAPIQ` to better reflect its purpose. The `TreeExplainer` should be the sole user-facing interface and the interventional treeexplainer is just the computation backend.

### Describe your proposed solution

Do the renaming in all occurrences (code, docs, comments).

### Impact

None

Advueu963:
With this change we should also addres #576

mmschlk:
Closed in #580

PR #580, opened 2026-08-20
Improving and Renaming InterventionalTreeExplainer

### LINKED ISSUE #576: Add STII Computation to InterventionalTreeExplainer
### Describe the workflow you want to enable

The InterventionalTreeExplorer currently does not support the computation of `STII`, which is totally doable as it requires simple arithmetic and even supports custom weight functions. The goal would be to not use the CUSTOM path, but to support STII natively via C++ weights.cpp

### Describe your proposed solution

The solution is simply adding STII into the general weight formula of the ProxySHAP [paper](https://arxiv.org/pdf/2605.22738) and then solving for a closed form solution. 

### Describe alternatives you've considered, if relevant

None

### Additional context

_No response_

### Impact

None

Advueu963:
## Motivation and Context

This PR closes #576 and #574 by adding (A) Support for STII any order in the InterventionalTreeExplainer and (B) renaming the latter to the InterventionalTreeSHAPIQ.

Alongside inconsistencies in the Interactionvalues.baseline_value was fixed.

## Public API Changes

-   [ ] No Public API changes

## How Has This Been Tested?

Local Unit tests.

## Checklist


src/shapiq/interaction_values.py
```diff
@@ -125,6 +125,7 @@ def __init__(
         )
 
         self.interactions = interactions
+        self.baseline_value = baseline_value
```
mmschlk: I am a bit scared of this one. Does this change anything for other indices or so in a bad way? I am just double checking since we already had these issues with baseline value setting in the past.

Advueu963: Yeah so I decided to revert this, because I am more and more realizing I am not happy with the way we are dealing with baseline values and (). So this change actuallly would not have affected anything from the approximator/explainer side, as they all already set ()==`baseline_value`. And the function `_update_interactions_for_index` would have just have had no effect. The change was done, results could have then had different entries for Banzhaf values () and `baseline_value` when they are not coming from any approximators/explainers. I then spotted that I am actually returning the updated one beforehand but do not update the `self.baseline_value` again. So yeah... I think I would like some consistency here. But I would like to think about this maybe as a seperate PR.

mmschlk (approved):
With the changes, this is now good to go (after Resolving the merge issue with main of course).

PR #583, opened 2026-08-23
Align LeverageSHAP with the paper's released implementation (functional pairing_trick, deterministic counts, exact budget)

rtealwitter:
## Motivation and Context

Follow-up to #524 (`LeverageSHAP`, Musco and Witter 2025), aligning the approximator with the paper and its released implementation. I'm the paper's co-author; while auditing the shapiq port against our reference code (https://github.com/rtealwitter/leverageshap) we found three faithfulness gaps, fixed here:

- **`pairing_trick` was silently ignored.** It is now functional: `True` (the default, as in the paper) forces each sampled coalition's complement into the sample; `False` reproduces the paper's "without paired sampling" ablation.
- **Per-size sample counts.** New `deterministic_counts=True` default fixes each coalition size's count to the expected value of the paper's Binomial draw (largest-remainder rounding), matching the paper's released implementation and reported experiments. `deterministic_counts=False` keeps the previous random Binomial draws (Algorithm 2 as stated in the paper). This changes which coalitions are sampled vs v1.6.0 — flagged in the CHANGELOG, and we're happy to flip the default if you prefer strict backward compatibility.
- **Budget accounting.** `estimation_budget` is now exact in the default mode, and `_find_c` no longer undershoots full enumeration at `budget == 2**n` (previously the largest layer was a Binomial draw with p ≈ 1 − 1e−12 instead of forced enumeration).

## Public API Changes

-   [ ] No Public API changes

New constructor parameter `deterministic_counts: bool = True`; `pairing_trick` now has an effect. Both documented in the class docstring and CHANGELOG. No signature removals.

## How Has This Been Tested?

- 320 unit tests for this approximator (from 251 on `main`), 100% line coverage of `leverageshap.py`.
- Mutation testing: 23 hand-crafted mutants of the sampling/budget logic; tests kill all of them.
- Head-to-head benchmark vs the paper's released implementation on the paper's 8-dataset grid (10 seeds × budgets 5n–160n): pooled error ratio 0.998, 95% bootstrap CI [0.935, 1.072] (parity), while running 3–5× faster than the reference code and avoiding its O(2^n) memory at large n.
- Exactness check: at `budget == 2**n` the estimate matches exact Shapley values to ~1e−15.

## Checklist

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Uoifa2WZoPhMsSsS7h8vAS

## Post-review alignment note

Three independent blind audits compared this PR with the released `leverageshap` code and paper pseudocode. Commit `e3083469` applies the resulting fixes: deterministic regression weights now use each layer’s realized fixed-count inclusion probability; the optional Binomial path draws the even middle layer from the full Binomial before halving; and exact NumPy Binomial draws are used throughout the int64 trial range.

The shapiq adapter intentionally rounds odd deterministic budgets down to preserve the library’s hard budget ceiling, whereas the released implementation rounds them up; largest-remainder ties may also choose a different valid stratum. The class documentation and CHANGELOG now state these differences explicitly.

Post-fix validation: all 325 LeverageSHAP tests pass, and `uv run pre-commit run --all-files` passes. The earlier head-to-head benchmark predates these final audit fixes and establishes aggregate parity for the earlier implementation rather than bit-for-bit validation of this commit.


src/shapiq/approximator/regression/leverageshap.py
```diff
@@ -90,24 +106,40 @@ def __init__(
...
+                (largest-remainder rounded so the total matches the budget exactly)
+                instead of drawing it at random -- this is what the paper's released
+                implementation and reported experiments use. If ``False``, draws
+                ``m_s`` at random as Algorithm 2 literally describes. See the class
+                docstring's Note for the accuracy-guarantee caveat.
+
             **kwargs: Additional keyword arguments (not used, only for compatibility).
         """
+        self.deterministic_counts = deterministic_counts
+        self.pairing_trick = pairing_trick
```
mmschlk, line 111: The super should have also a pairing variable. Setting it here again is not so nice imo.

mmschlk: Ah needs to be there as the super grabs it from the sampler, which we are now omitting. I will defer a rewrite of this to some other time.

mmschlk:
Okay nice! Yes this deterministic setting is a nice addition. We were our own not sure which "approach" would be the correct one. @rtealwitter, I made some changes. Particularly how the sampler is making sure the budget is "exactly" allocated as before you had quite a hefty code comment but no real guard. Yes empirically this is okay but the `assert` in the code was a no-go for me as a) the assert would crash a user's run and worse b) asserts are not evaluated in `python -O` runs which is the default in some prod envs.

Could you re-check if the current state is fine for you? :)

mmschlk:
One note. LeverageSHAP never shipped. It's gonna ship with 1.7.0. No released version of shapiq carried LeverageSHAP.

mmschlk (approved):
This is good to go from my side. But letting it sit for a day waiting for @rtealwitter to come back.

mmschlk:
Aaaah okay, I see what you did there. Now I understand what you do with your C sampling trick. Then, I think this is good to go for now (will do a retouching of the approximators for v2 anyways but this makes leverages hap behave exactly like the paper, which is good).

PR #586, opened 2026-08-24
fix(tree): validate_tree_model returns caller-owned trees (no more TreeModel mutation)

### LINKED ISSUE #544: LinearTreeSHAP mutates the caller-owned TreeModel in place (feature ids remapped under other explainers)
### What happens

Constructing a `LinearTreeSHAP` explainer silently mutates the `TreeModel` you pass in. After the call, the model's feature indices have been overwritten in place, so any code that keeps a reference to that `TreeModel` — to inspect which features the tree uses, to map attributions back to original feature names, or to reuse it with another explainer — reads corrupted metadata.

### Reproduction

```python
import numpy as np
from sklearn.tree import DecisionTreeRegressor
from shapiq.tree.linear import LinearTreeSHAP
from shapiq.tree.validation import validate_tree_model

rng = np.random.default_rng(0)
X = rng.normal(size=(500, 8))
y = 10 * X[:, 5] - 7 * X[:, 7]          # only features 5 and 7 carry signal
model = DecisionTreeRegressor(max_depth=4, random_state=0).fit(X, y)

tree = validate_tree_model(model, class_label=None)[0]
print("features the tree splits on:", sorted(set(tree.features[tree.features >= 0])))
# -> [5, 7]   (correct: the original feature ids)

LinearTreeSHAP(tree)   # just constructing the explainer

print("features the tree splits on:", sorted(set(tree.features[tree.features >= 0])))
# -> [0, 1]   (WRONG: the caller's TreeModel was remapped in place)
```

The user never asked to modify `tree`; merely constructing an explainer over it rewrote its `features` from the real ids `[5, 7]` to internal compressed ids `[0, 1]`.

### Why it matters

Constructing an estimator must not modify the object passed to it (cf. scikit-learn's `check_estimators_do_not_mutate`). Here it does, so a `TreeModel` shared across explainers, or simply inspected afterwards, yields wrong feature ids. `TreeSHAPIQ` already guards against exactly this — it deep-copies the validated model before `reduce_feature_complexity()` — so the two explainers behave inconsistently on the same input.

### Root cause

`validate_tree_model` returns `TreeModel` inputs **by reference**, and `LinearTreeSHAP.__init__` then calls `reduce_feature_complexity()` on the result, which remaps `features` / `feature_ids` / `n_features_in_tree` **in place** on the caller's object.

### Proposed fix

Deepcopy the validated tree before reducing — a one-line change in `src/shapiq/tree/linear/explainer.py`, matching what `TreeSHAPIQ` already does. PR with a regression test follows.

### Environment

shapiq current `main`, Python 3.14, NumPy 2.x (the aliasing is structural, version-independent).

mmschlk:
Fixes #544. Supersedes #546 with a boundary-level fix instead of a per-call-site deepcopy.

### Problem

`validate_tree_model` returned `TreeModel` and list-of-`TreeModel` inputs by reference, and `LinearTreeSHAP.__init__` then called `reduce_feature_complexity()` on the result — rewriting the caller's `features` / `feature_ids` / `n_features_in_tree` in place. Since `TreeExplainer` routes `index="SV", max_order=1` to `LinearTreeSHAP`, merely constructing an explainer over a caller-owned `TreeModel` corrupted it.

Beyond the metadata corruption reported in #544, this produced **silently wrong attributions**: `InterventionalTreeSHAPIQ` reads `tree.features` raw (it does not consult `feature_map_internal_original`), so a tree previously touched by `LinearTreeSHAP` yields attributions on the wrong features:

```
clean tree:            {5: 4.534, 7: -6.964}
after LinearTreeSHAP:  {0: -4.177, 1: 0.661}
```

`TreeSHAPIQ` already deep-copied before reducing; `LinearTreeSHAP` was the inconsistent one, and `TreeExplainer`'s own `self._trees` were aliased into its `LinearTreeSHAP` children as well.

### Fix

- `validate_tree_model` now returns trees the caller owns: `TreeModel` and list-of-`TreeModel` inputs are deep-copied (dict and converted library models already produced fresh objects). One invariant makes every current and future consumer mutation-safe at the boundary, including `TreeExplainer`'s internal aliasing.
- The now-redundant `copy.deepcopy` in `TreeSHAPIQ.__init__` is removed — it was also wasted work on freshly-converted models.
- A list input containing non-`TreeModel` elements now raises `TypeError` instead of silently returning an empty list.

### Performance

Timed both states (stash-interleaved, medians):

- **ProxySHAP + XGBoost readout (`_extract_tree`): unchanged** — it passes the raw fitted model, which goes through the untouched conversion branch (1.47 ms before vs 1.39–1.57 ms across after-runs, pure noise; `validate` best-case identical at 0.635 ms).
- All library-model paths (sklearn / XGBoost / LightGBM / CatBoost): no new copies; `TreeSHAPIQ` on converted models drops one deepcopy.
- Caller-supplied `TreeModel` inputs pay the copy correctness requires: ~26 µs per tree, once, at construction (~2.6 ms for a 100-tree list; ~3% on interventional-explainer construction, invisible next to a single `explain` call). Per-explanation costs are unchanged everywhere.

### Testing

- New tests in `test_tree_explainer_validate.py`: `validate_tree_model` returns owned copies for both `TreeModel` and list inputs; mixed lists raise `TypeError`.
- New regression tests in `test_tree_bugfix.py`: `LinearTreeSHAP` construction leaves the caller's `TreeModel` untouched, and the `TreeExplainer` SV fast path no longer corrupts a tree that is subsequently explained interventionally (values match a clean tree). All four verified red without the fix.
- Full `tests_tree_explainer` suite (271 passed), ProxySHAP approximator/dispatch/integration suites, `ruff`, `ruff-format`, and `ty` all pass.


Advueu963 (approved):
LGTM!

Issue #585, opened 2026-08-23
RegressionMSR/ProxySHAP with index="BV" and a linear proxy raises TypeError, dependent on per-process state


rtealwitter:
While benchmarking `RegressionMSR` for #584 we hit a pre-existing bug that is independent of that PR (identical with and without it, verified at the same commit with the PR's changes removed).

**Symptom.** `RegressionMSR(index="BV", proxy_model="linear")` (equivalently `ProxySHAP` routed through the linear extraction with a Banzhaf index) raises on `approximate()`:

```
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

**What we observed.**
- In two full benchmark grids (27 (n, budget) points × 5 seeds, run at `a19c8b74`), *every one* of the 135 BV+linear cases raised this error, while zero SV+linear and zero BV+xgboost cases did.
- However, three isolated reproduction attempts — a single-case replay with identical parameters and seeds, a same-node replay, and a replay of the entire 540-case loop in the same order up to the failing case — all succeeded. Same interpreter, same env, same node in two of the attempts.
- Our best hypothesis is per-process nondeterminism unrelated to `random_state` — e.g. Python's per-process hash randomization (`PYTHONHASHSEED` unset) feeding a dict/set-ordering-dependent step in the linear+BV extraction route (`approximator/proxy/_routes.py`) — but we did not pin it down further.

So: deterministic *within* a process (all 135 cases in an affected process fail identically), but dependent on some per-process state. If it is hash randomization, running the suite with several different `PYTHONHASHSEED` values should flush it out.

Happy to share the repro scripts and full case list if useful.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Uoifa2WZoPhMsSsS7h8vAS

mmschlk:
Thanks @rtealwitter. I also have seen your PR. @Advueu963 and I are working on a different fix to your PR. This should tackle your bug as well.

mmschlk:
Closed with #587.

PR #584, opened 2026-08-23
RegressionMSR: closed-form MSR correction, paper's sampling kernel as default, fit diagnostics

rtealwitter:
## Motivation and Context

Follow-up to #519 (`RegressionMSR`, Witter et al. 2025). I'm a co-author of the paper; while auditing the shapiq port against our reference implementation (https://github.com/rtealwitter/regressionMSR) we made four improvements, fixed here:

- **Closed-form MSR adjustment.** The `"msr"` adjustment is now a closed-form Horvitz–Thompson correction on the coalitions already sampled, instead of re-sampling an independent `CoalitionSampler` and running a second Monte Carlo pass on the residual game. Numerically identical to the previous path (≤1e−10) and removes the slowest step of `approximate()`. Indices other than `"SV"`/`"BV"` delegate the whole call to `ProxySHAP.approximate()` unchanged.
- **Paper's sampling kernel as the default `sampling_weights`.** The previous generic bowl-shaped default was not the sampling scheme the paper uses; the new default reproduces the reference's kernel (converted from its per-coalition density to shapiq's per-size mass convention, computed in log-space so it is stable to n ≥ 1000), including the reference's separate Leverage-SHAP kernel for linear proxies with `index="SV"`. An explicit `sampling_weights=` still overrides.
- **Diagnostics.** New public attributes `train_residual_ratio` and `correction_norm` (set by a fast-path `approximate()`, else `None`) expose how well the proxy fits and how much correction survives — useful because strong proxies (e.g. XGBoost) often interpolate the sampled coalitions.
- **CHANGELOG wording.** The v1.5.0 entry calls the msr/svarm/kernel adjustments "unbiased"; the k=1 all-samples variant implemented here is consistent, not unbiased (the paper's unbiased variant needs k ≥ 2 cross-fitting). Corrected.

## Public API Changes

-   [ ] No Public API changes
-   [x] Yes, Public API changes (Details below)

Behavior change: the default `sampling_weights` for `index in {"SV", "BV"}` (documented in the class docstring and CHANGELOG; explicit `sampling_weights=` overrides, other indices unchanged). New public attributes `train_residual_ratio`, `correction_norm`. No signature removals.

## How Has This Been Tested?

- 116 unit tests for this approximator (from ~5 exercising it on `main`), 100% line coverage of `regressionmsr.py`; `ProxySHAP` tests unchanged and passing.
[...]

PR #587, opened 2026-08-24
Improving ProxySHAP and RegressionMSR

Advueu963:
## Motivation and Context

This PR was motivated by PR #584, which revealed inconsistencies between the implemented RegressionMSR and the paper's code. Yet the original PR was heavily loaded with extensive docstrings and had significantly increased the number of code lines in `RegressionMSR`. We tackle both issues by refactoring `ProxySHAP`, identifying closed-form solutions to the original paper weights. Additionally speedups are now achieved, by dropping `"msr"` / `"svarm"` / `"kernel"` adjustments, which each instantiated a *second* approximator with its own `CoalitionSampler` and ran a second Monte Carlo pass over a `ResidualGame`, which forced `ProxySHAP` to override `random_state` with a fixed seed so both samplers would draw identical coalitions.

The adjustment is now a self-contained, vectorized MSR routine that estimates all interactions in a single matrix product over the coalitions already sampled. 
This enabled high-dimensional runtime drops by more than an order of magnitude: 12× at 1617 features with a budget of 100 coalitions.
grafik

Additionally, we have fixed a bug in the `BII` path using a linear surrogate, by enabling the MoebiusConverter to compute `BII` (see #585). 
Also, ProxySHAP now supports the `STII` index, and the `RegressionMSR` sampling kernels are aligned with Witter et al. (2025).

## Public API Changes

-   [ ] No Public API changes

**`ProxySHAP`**
- `adjustment` (`"none"`/`"msr"`/`"svarm"`/`"kernel"`) is replaced by `apply_msr_adjustment: bool = False`; the `"svarm"` and `"kernel"` variants are no longer available.
- New `kfolds: int = 1` for out-of-fold proxy residuals. By default, it is set to 1, as it yields the best estimates Witter et al., 2025; Thies et al., 2026).
- `random_state=None` is no longer silently replaced by a fixed seed.
- Default `sampling_weights` is now uniform by size, not `Approximator`'s KernelSHAP-style default.
- Now computes `BII` and `STII`, and is registered for both in `setup_approximator`.

**`RegressionMSR`**
- Default `sampling_weights` is now the paper's kernel: `binom(n, s)` for `BV`, uniform (Leverage SHAP) for `SV` with a linear proxy, `1 / sqrt(s * (n - s))` otherwise. An explicit array still overrides.
- An invalid `index` now raises at construction.

**`MoebiusConverter`**
- Supports `BII`. `BV` now routes through the base-interaction routine, which fixes `estimated`/`estimation_budget` being dropped for a linear proxy, but also changes `min_order` from 0 to 1 and drops the `()` entry.
- Side effect: `ValidSparseIndices` / `ValidProxySPEXIndices` (aliases of the converter's `Literal`) now include `BII`.

## How Has This Been Tested?

- Full unit suite: 1346 passed. 
- New and updated tests (66 across the ProxySHAP suites): full-budget exactness for `SV`/`SII`/`STII`/`BII`/`BV`, lazy interaction lookup at n=1778, and MSR support-restriction consistency.
- Equivalence: the new routine matches `SHAPIQ.monte_carlo_routine` on identical residuals to ≤4.4e-16; the `RegressionMSR` kernels match a direct evaluation of the paper's kernel to ≤3.4e-14 up to n=1776.

## Checklist


mmschlk:
Fixed #585

mmschlk:
Hahah. I love the „projected 84x improvements“ 😂😂😂

mmschlk:
This PR is related to #584

Advueu963:
# Sampling weights in `RegressionMSR`: derivation

`RegressionMSR` implements the Regression-MSR estimator of Witter, Liu and Musco (2025),
[*Regression-adjusted Monte Carlo Estimators for Shapley Values and Probabilistic Values*](https://arxiv.org/abs/2506.11849),
whose coalition-sampling distribution is given as a closed-form kernel. This note derives what
that kernel evaluates to for the two indices `RegressionMSR` supports, `"SV"` and `"BV"`. In every
case it collapses to an elementary expression, which is why `RegressionMSR.__init__` builds its
`sampling_weights` as three short NumPy expressions instead of evaluating the kernel.

## Notation

A probabilistic value with coefficients $p_0, \dots, p_{n-1}$ is

$$\phi_i(v) \;=\; \sum_{S \subseteq [n] \setminus \{i\}} p_{|S|}\,\bigl[v(S \cup \{i\}) - v(S)\bigr],
\qquad
p_k^{\mathrm{SV}} = \frac{k!\,(n-k-1)!}{n!} = \frac{1}{n\binom{n-1}{k}},
\qquad
p_k^{\mathrm{BV}} = \frac{1}{2^{\,n-1}} .$$

**The paper's kernel.** Each *individual coalition* $S$ is sampled with probability proportional to

$$D(s) \;=\; \sqrt{\,p_{s}^{2}\left(1 - \frac{s}{n}\right) \;+\; p_{s-1}^{2}\,\frac{s}{n}\,},
\qquad s = |S|.$$

The prefactors make the edges self-handling: at $s = 0$ the non-existent $p_{-1}$ term is killed by
$s/n = 0$, and at $s = n$ the non-existent $p_n$ term is killed by $1 - s/n = 0$.

**shapiq's convention.** `sampling_weights[s]` is a *per-size* probability mass — the probability
that a drawn coalition has size $s$ — not a per-coalition density. With $\binom{n}{s}$ coalitions
of size $s$,

$$\mathrm{mass}[s] \;\propto\; D(s)\,\binom{n}{s}.$$

This conversion is the crux: that $\binom{n}{s}$ cancels the binomials hidden inside $p_k$, which
is why all three cases simplify.

**Two identities**, from $\binom{n-1}{s} = \binom{n}{s}\frac{n-s}{n}$ and
$\binom{n-1}{s-1} = \binom{n}{s}\frac{s}{n}$:

$$p_s^{\mathrm{SV}} = \frac{1}{\binom{n}{s}(n-s)},
\qquad\qquad
p_{s-1}^{\mathrm{SV}} = \frac{1}{\binom{n}{s}\,s}.$$

## Case 1 — Banzhaf (`index="BV"`)

$p_k^{\mathrm{BV}} = p$ does not depend on $k$, so the two prefactors sum to one and

$$D(s)^2 \;=\; p^2\left(1 - \frac{s}{n}\right) + p^2\,\frac{s}{n} \;=\; p^2$$

for every $s$, ends included. $D$ is constant, so

$$\boxed{\ \mathrm{mass}[s] \;\propto\; \binom{n}{s}\ }$$

the binomial distribution $\mathrm{Binom}(n, \tfrac12)$ — each player independently in or out with
probability $\tfrac12$, which is what Banzhaf means.

## Case 2 — Shapley with a non-linear proxy

Substituting the two identities, for $1 \le s \le n-1$:

$$D(s)^2 \;=\; \frac{1}{\binom{n}{s}^2 (n-s)^2}\cdot\frac{n-s}{n} + \frac{1}{\binom{n}{s}^2 s^2}\cdot\frac{s}{n}
\;=\; \frac{1}{n\binom{n}{s}^{2}}\left(\frac{1}{n-s} + \frac{1}{s}\right)
\;=\; \frac{1}{s(n-s)\binom{n}{s}^{2}},$$

so $D(s) = \bigl[\binom{n}{s}\sqrt{s(n-s)}\bigr]^{-1}$ and the binomial cancels on conversion:

$$\boxed{\ \mathrm{mass}[s] \;=\; \frac{1}{\sqrt{s(n-s)}}\ } \qquad (1 \le s \le n-1).$$

At the ends one prefactor vanishes, leaving a value that is *not* the continuous extension of the
interior formula. With $p_0 = p_{n-1} = 1/n$:

$$D(0)^2 = p_0^2 = \frac{1}{n^2}, \qquad \mathrm{mass}[0] = D(0)\binom{n}{0} = \frac{1}{n},$$

and symmetrically $\mathrm{mass}[n] = 1/n$.

## Case 3 — Shapley with a linear proxy: Leverage SHAP

Here the reference switches to the Leverage SHAP density of Musco and Witter (2025),
[*Provably Accurate Shapley Value Estimation via Leverage Score Sampling*](https://arxiv.org/abs/2410.01917):

$$D(s) = (p_s + p_{s-1})\, s\,(n-s) \quad (1 \le s \le n-1), \qquad D(0) = D(n) = 0.$$

Using the same identities, $p_s + p_{s-1} = \dfrac{n}{s(n-s)\binom{n}{s}}$, so the $s(n-s)$ factor
cancels exactly:

$$D(s) \;=\; \frac{n}{\binom{n}{s}}, \qquad\qquad
\boxed{\ \mathrm{mass}[s] \;=\; D(s)\binom{n}{s} \;=\; n \;=\; \text{const}\ }$$

i.e. uniform over sizes $s \in \{1, \dots, n-1\}$.

This reproduces the Leverage SHAP paper rather than coinciding with it: its Lemma 3.2 gives the
leverage score of the row indexed by $z$ with $0 < \lVert z\rVert_1 < n$ as
$\ell_z = \binom{n}{\lVert z \rVert_1}^{-1}$, exactly the density $D(s)$ above up to the constant
$n$. The paper states the consequence directly — *"leverage score sampling ensures that all subset
sizes are equally represented"* — and describes the procedure as: draw a size $s$ uniformly from
$\{1,\dots,n-1\}$, then a coalition of that size uniformly at random.

## Summary

| `index` | proxy | per-size mass | distribution |
|---|---|---|---|
| `"BV"` | any | $\binom{n}{s}$ | $\mathrm{Binom}(n, \tfrac12)$ |
| `"SV"` | linear | $\text{const}$ | uniform over sizes (Leverage SHAP) |
| `"SV"` | non-linear | $1/\sqrt{s(n-s)}$, ends $1/n$ | $\sqrt{\ }$ of the KernelSHAP kernel |

The Shapley-side kernels form a one-parameter family in the exponent of $s(n-s)$, since shapiq's
default KernelSHAP-style weights are $\propto 1/(s(n-s))$:

$$\mathrm{mass}[s] \;\propto\; \bigl(s(n-s)\bigr)^{-\alpha},
\qquad \alpha = 0 \ \text{(Leverage SHAP)}, \quad
\alpha = \tfrac12 \ \text{(this kernel)}, \quad
\alpha = 1 \ \text{(KernelSHAP)}.$$

## Verification

Closed forms versus a direct log-space evaluation of $D(s)$ (per-index $p_k$, a log-sum-exp over
the two terms, then $\times\binom{n}{s}$ and normalization). Maximum absolute deviation of the
normalized weight vectors:

| $n$ | 2 | 3 | 5 | 12 | 101 | 1000 | 1776 |
|---|---|---|---|---|---|---|---|
| `"SV"` | 1.1e-16 | 5.6e-17 | 1.7e-16 | 1.7e-16 | 6.1e-16 | 3.5e-15 | 3.0e-15 |
| `"BV"` | 5.6e-17 | 1.7e-16 | 1.7e-16 | 3.3e-16 | 2.9e-15 | 1.2e-14 | 3.4e-14 |

The Banzhaf branch additionally matches $\binom{8}{s}/2^8 = [1,8,28,56,70,56,28,8,1]/256$ at
$n = 8$ to $1.1\times10^{-16}$. The Leverage SHAP branch is passed as `np.ones(n + 1)` rather than
as zero-at-the-ends uniform weights; the two are different vectors but induce the same sampler
(shapiq's border trick enumerates sizes $0$ and $n$ regardless of their weight), verified across
$n \in \{8, 12, 20, 170, 500\}$ and budgets $\{30, 200\}$ as identical coalition matrices and
identical `log_sampling_adjustment_weights` (max deviation $8.9\times10^{-16}$).

src/shapiq/approximator/proxy/proxyshap.py
```diff
@@ -59,7 +186,8 @@ def __init__(
         index: ValidProxySHAPIndices = "k-SII",
         proxy_model: ProxyModel | ProxyModelWithHPO | ProxyLiteral = "xgboost",
         hpo: bool = False,
-        adjustment: str = "msr",
+        kfolds: int = 1,
+        apply_msr_adjustment: bool = False,
```
mmschlk: I don't like the param name: I think `adjustment: bool = False` for ProxySHAP and `True` for RegressionMSR is better. Also I would move it up before kfolds again as it's more important. Also I would name it `k_folds`.

src/shapiq/approximator/proxy/proxyshap.py
```diff
@@ -71,38 +199,47 @@ def __init__(
...
             hpo: If ``True``, wrap a string-resolved gradient-boosting proxy (``"xgboost"`` /
                 ``"lightgbm"``) in its default grid search (the HPO-informed proxy). Defaults to
                 ``False`` (a bare estimator). Has no effect when ``proxy_model`` is a passed-in
                 estimator/wrapper, or for the ``"tree"`` / ``"linear"`` tags.
-            adjustment: Method for adjusting the proxy model's predictions to better match the true value function. Options are "none" (no adjustment), "msr","svarm" (statified MSR), "kernel" (KernelSHAPIQ).
+            kfolds: Number of folds the sampled coalitions are split into. With
+                the default ``1``, a single proxy is fit on all sampled coalitions and its
+                residuals are computed in-sample. For values ``> 1``, one proxy is fit per fold
+                (KFold) on the training split, its interactions are extracted, and its residuals
+                are computed on the held-out split only; the per-fold results are averaged.
```
mmschlk: add info about scaling. That k fold is making it slow. Maybe add the notion that they are getting averaged. And maybe say that default 1 is also reported in the papers.

mmschlk: I changed this in 9c12d44

src/shapiq/approximator/proxy/proxyshap.py
mmschlk: I changed this in 9c12d44

mmschlk (approved):
Okay I went through this and the improvements are as intended!

PR #588, opened 2026-08-24
Add Quadrature-TreeSHAP and fix numerical instability in path-dependent TreeSHAP

### LINKED ISSUE #545: TreeSHAP-IQ Vandermonde solves: silent wrong values for interpolation degrees ~32-59, LinAlgError crash at ~60+
### What happens

Explaining a moderately deep tree with `LinearTreeSHAP` returns Shapley values that do not sum to the model's prediction — the efficiency property is violated by many orders of magnitude — and for very deep trees it crashes outright with an unexplained `LinAlgError`. There is no warning: the attributions look like ordinary numbers and are silently wrong.

### Reproduction

A plain sklearn regression tree on data where deep splits arise naturally (here: sparse / one-hot-style indicator features):

```python
import numpy as np
from sklearn.tree import DecisionTreeRegressor
from shapiq.tree.linear import LinearTreeSHAP

rng = np.random.default_rng(7)
X = (rng.random((20000, 100)) < 0.02).astype(float)   # rare indicator features
w = rng.normal(scale=10.0, size=100)
y = X @ w + rng.normal(size=20000)

for max_depth in (30, 40, 55, None):
    model = DecisionTreeRegressor(max_depth=max_depth, random_state=0).fit(X, y)
    x = X[0]
    pred = model.predict(x.reshape(1, -1))[0]
    sv = LinearTreeSHAP(model).explain_function(x)
    print(f"depth={model.get_depth():3d}  prediction={pred:8.3f}  "
          f"efficiency error = sum(SV) - prediction = {sv.values.sum() - pred:12.4g}")
```

Output on current `main`:

```
depth= 30  prediction= -13.218  efficiency error = sum(SV) - prediction =       -1.677
depth= 40  prediction= -13.673  efficiency error = sum(SV) - prediction =   -1.344e+05
depth= 55  prediction=  -7.949  efficiency error = sum(SV) - prediction =    -2.45e+12
depth=100  prediction=  -7.949  efficiency error = sum(SV) - prediction =     2.08e+29   # max_depth=None
```

The Shapley values should sum to the prediction (efficiency error ≈ 0). Instead the error grows without bound as the tree deepens, and at full depth a tree trained with the **default** `max_depth=None` returns complete garbage — with no warning. Push the depth past ~60 and you get `numpy.linalg.LinAlgError: Singular matrix` from inside the explainer instead.

### Root cause

`TreeSHAPIQ` and `LinearTreeSHAP` build their interpolation N-matrices by explicitly inverting Vandermonde systems over prefixes of the interpolation grid, `inv(np.vander(D[:i]).T)`. These systems are severely ill-conditioned, and the conditioning is **non-monotonic in the prefix length** (it peaks at interior prefixes, `i ≈ n/2`, not at the full grid). Measured on the default Chebyshev grids there are three failure bands as the interpolation degree grows (degree = `min(depth, n_features_in_tree)` for `TreeSHAPIQ`, the tree depth for `LinearTreeSHAP`):

| degree | peak prefix cond. number | behaviour today |
|---|---|---|
| ~27–31 | `1e12`–`1e14` | silent precision loss |
| ~32–59 | rank-deficient to machine precision | `inv` returns garbage → silently wrong values |
| ~60+ | singular | `LinAlgError` crash |

### Proposed fix

Drop the explicit inverse and centralise the solve: certify each interpolation grid by measuring its peak prefix condition number once (cached), take a plain `np.linalg.solve` for certified grids, and route the rest through one SVD (`lstsq`) that detects ill-conditioning / rank-deficiency, degrades to least squares instead of crashing, and emits a single coalesced `RuntimeWarning` per N-matrix stating when the values are unreliable. PR with tests follows.

### Environment

shapiq current `main`, Python 3.14, NumPy 2.x (condition numbers reproduce across platforms to within rounding).

mmschlk:
## Summary

Fixes #545. Path-dependent tree explanations get a numerically exact, faster algorithm — and it is now the default.

1. **`QuadratureTreeSHAP` — the new default path-dependent algorithm.** Implements Quadrature-TreeSHAP (Wettenstein, Mitchell, Yu, 2026; first order equals TreeGrad): Shapley values and any-order Shapley interactions are computed as Gauss-Legendre integrals of weighted-Banzhaf interaction polynomials. In the homogenized parameterization used here every maintained factor lies in (0, 1], so the computation is exact in float64 at **any** tree depth and any number of features per path. One algorithm covers `SV`, `SII`, `k-SII` at any order, plus `BV`/`BII` natively (Banzhaf = a single polynomial evaluation at p = 1/2 — previously woodelf-only in path-dependent mode). The C extension is required (a missing extension is a packaging error and raises at construction; the numpy reference implementation lives in the test suite as the differential oracle); ensembles run as **one batched kernel call** over a shared feature space, with no per-tree Python objects on the hot path. The interaction output is **sparse**: the returned `InteractionValues` enumerates exactly the interactions whose features co-occur on at least one decision path (all others are structurally zero and absent), so wide models never materialize the dense `C(F, order)` enumeration — order-3 interactions of a 1000-feature forest are 89 co-occurring subsets computed in 0.28 ms where a dense output would need 1.7e8 slots. Order 1 stays a dense per-feature block with unchanged Shapley hot-path performance; interaction orders pay 1.6-2.2x for the sparse lookup (see the benchmark comment). `TreeExplainer` in `"pathdependent"` mode now always builds this explainer — the previous order- and size-based routing rules between LinearTreeSHAP, TreeSHAP-IQ, and Woodelf are gone. Woodelf remains the auto fast path for large *interventional* inputs and can still be forced with `backend="woodelf"`.

2. **Precision guards for the polynomial explainers.** `LinearTreeSHAP` and `TreeSHAPIQ` stay available standalone but lose precision as the number of *distinct features along a single root-to-leaf path* grows (~10x error per +2 features, independent of how the interpolation systems are solved) and previously returned silently wrong values or crashed with `LinAlgError`. Both now measure this quantity at construction: from 24 features per path they emit a `TreeNumericalPrecisionWarning` stating the expected error magnitude, from 30 they refuse with a `TreeNumericalPrecisionError` pointing at the quadrature explainer. Two accompanying fixes remove all crashes for legitimately computable trees: `TreeSHAPIQ` sizes its interpolation grid by the maximum number of distinct features per path (instead of `min(depth, n_features)`), and `LinearTreeSHAP` no longer constructs N-matrix rows its kernel never reads. Deep trees over few features per path now work at machine precision (a depth-60 tree over 8 features went from `LinAlgError` to an efficiency error of 9e-15).

3. **Harmonized, self-contained algorithms.** `LinearTreeSHAP`, `TreeSHAPIQ`, `QuadratureTreeSHAP`, and `InterventionalTreeSHAPIQ` all validate their own models, accept single trees and ensembles (aggregating internally), take a `class_index`, and share the `explain(x)` entry point — each usable independently of `TreeExplainer`, which simply instantiates one explainer per mode.

4. **Routing-fidelity bugfix (float32) and float64 interventional kernels.** XGBoost and CatBoost cast prediction inputs to float32 before routing; the converted trees now reproduce that cast (`TreeModel.input_precision`/`cast_input`), fixing instances that previously routed to a different leaf than the model's own prediction (on a measured XGBoost `hist` ensemble roughly half the instances routed differently somewhere). LightGBM compares in float64 and is untouched; sklearn's thresholds are exact float32 midpoints, making its internal cast unobservable. The interventional C kernels moved from float32 to float64 end to end, and their three unused entry points were deleted.

## Correctness

Numpy and C implementations agree to <= 9e-15 everywhere. Both were validated against:

- `TreeSHAPIQ` (orders 1-3, `SV`/`SII`/`k-SII`) and `LinearTreeSHAP` on shallow trees across model families: sklearn regressors/classifiers, XGBoost (incl. categorical splits and NaN missing-value routing), LightGBM, GradientBoosting, HistGradientBoosting, CatBoost — agreement <= 1e-8. CatBoost's oblivious trees contain unreachable zero-cover leaves whose edges produced 0/0 in the quadrature factors; they now clamp to the exact c -> 0 limit.
- Exact brute-force Shapley, Shapley-interaction, and Banzhaf-interaction enumeration on small trees (<= 1e-12).
- An independent per-leaf oracle on depth-40 sparse trees where the polynomial explainers cannot run (<= 9e-16); efficiency holds to 1e-14 up to depth 100.
- Results are invariant under adding quadrature points, confirming the default `ceil(d/2)`-point rule is exact.

## Runtime (single explanation point, median wall-clock, Apple Silicon)

Tree A: depth 6, 64 leaves, 8 features. Tree B: depth 16, 19,615 leaves, 50 features. Tree C: depth 40, 3,754 leaves, 100 features (the regime of #545).

| tree | task | LinearTreeSHAP (C++) | TreeSHAP-IQ (py) | Quadrature (numpy) | Quadrature (C++) |
|---|---|---|---|---|---|
| A | SV | 17 us | 2.5 ms | 790 us | **7 us** |
| A | SII order 2 | – | 5.9 ms | 1.2 ms | **13 us** |
| A | SII order 3 | – | 9.0 ms | 1.5 ms | **23 us** |
| B | SV | 3.2 ms | 737 ms | 226 ms | **1.7 ms** |
| B | SII order 2 | – | 2.06 s | 625 ms | **6.1 ms** |
| B | SII order 3 | – | 14.8 s | 2.9 s | **30 ms** |
| C | SV | refused | refused | 44 ms | **0.44 ms** |
| C | SII order 2 | – | – | 221 ms | **3.7 ms** |
| C | SII order 3 | – | – | 3.3 s | **70 ms** |

- ~2x faster than the LinearTreeSHAP C++ kernel for Shapley values (half the grid points, no Offset/Norm arithmetic, no Vandermonde solves); 340-490x faster than TreeSHAP-IQ for interactions (complexity drops from O(L·D·C(F,s)) to O(N·d^(s-1))).
- Ensembles via the batched kernel: a 200-tree heloc random forest explains one instance in 2.8 ms vs 9.9 ms for the previous per-tree backend, with per-point cost constant in batch size.
- Against XGBoost's own native TreeSHAP (single thread): at parity or faster than 3.2.0's classic implementation (0.55-0.93x of its time); 1.2-2.4x slower than 3.3.0's new float32-SIMD Quadrature-TreeSHAP — the price of exact float64 with per-depth exact point counts.
- Construction cost 0.4-11 ms per tree, paid once per model. The C extension (30-500x faster than the numpy reference) is required; the numpy implementation lives in the test suite as the differential oracle.

Full sweeps (depth, features per path, ensemble size, dataset width, interaction order, batch size, XGBoost comparison) are in the benchmark comment below.

## Review

The change went through two independent full reviews (algorithm/numerics and C++/memory-safety/API lenses, both with randomized differential testing against brute-force oracles) plus an eight-angle review pass. Ten findings were confirmed and fixed, most notably:

- **Feature-id ordering bug (pre-existing, also in `TreeSHAPIQ`):** raw `set` iteration over `feature_ids` can be non-ascending (e.g. `{6, 14}` iterates `[14, 6]`), which paired the packed value arrays with the *sorted* interaction lookup and silently permuted attributions between features (efficiency still held, so nothing downstream caught it; `LinearTreeSHAP` was unaffected, so backends disagreed). Fixed at the root in `TreeModel.reduce_feature_complexity` and in all three explainers; regression test added.
- `TreeExplainer` marked itself initialized before the underlying constructors ran, so a construction error turned into broken state on retry; the flag now moves after successful construction.
- The C binding validates all array dimensions/lengths and the output width before invoking the kernel (malformed arguments previously corrupted memory silently).
- The cibuildwheel smoke test and the native-extension-isolation suite now cover the new extension; `explain_X(n_jobs=...)` pre-builds the explainer so joblib workers stop re-constructing it per task.

Left as a follow-up: the `goes_left` routing logic exists in three C kernels and the constructor prologue in three classes; consolidation only makes sense once the polynomial explainers are retired from the package entirely.

## Notes

- Values computed by the polynomial explainers can shift at their (documented) round-off level for trees where the interpolation grid shrank; all used N-matrix rows are constructed identically.
- The guard thresholds (warn at 24, refuse at 30 distinct features per path) were calibrated on the reproduction of #545; both are module constants in `shapiq/tree/precision.py`.


mmschlk:
## Runtime benchmarks

🤖 _Benchmark generated with Claude Code._

---

Single-explanation runtime on TabArena datasets (superconductivity: 17k×81, regression; heloc: 8.4k×23, classification; bioresponse: 3k×1776 sparse binary). Median of ≤5 runs, warm-up excluded, Apple silicon (arm64), single thread. Benchmark code kept out of the PR by design; raw results available on request.

*Tables re-measured 2026-08-26 on the final (sparse-output) build; plots are from the earlier dense-output run — SV panels are unaffected, interaction timings shifted by the sparse-lookup factor in section e.*

### a · Tree depth (single tree, superconductivity, one instance)

image

| depth | feats/path | leaves | LinearTreeSHAP C++ | TreeSHAP-IQ py | Quadrature numpy | **Quadrature C++** |
|---|---|---|---|---|---|---|
| 4 | 4 | 16 | 0.02 ms | 0.4 ms | 0.2 ms | **0.01 ms** |
| 8 | 8 | 228 | 0.06 ms | 7.7 ms | 2.9 ms | **0.03 ms** |
| 16 | 16 | 5449 | 0.97 ms | 210.2 ms | 68.3 ms | **0.44 ms** |
| 24 | 24 | 12960 | 2.72 ms ⚠ | 473.3 ms ⚠ | 174.7 ms | **1.21 ms** |
| None | 29 | 14480 | 3.16 ms ⚠ | 586.1 ms ⚠ | 196.1 ms | **1.61 ms** |

⚠ = degrading precision band (≥24 features per path). Order-2 interactions at full depth: **16.1 ms** (quadrature C++) vs 2.7 s (TreeSHAP-IQ, explain-only), ~165×.

### a2 · Features per path = depth (synthetic sparse indicators, issue #545 regime)

image

| depth = feats/path | leaves | LinearTreeSHAP C++ | TreeSHAP-IQ py | **Quadrature C++** | quad efficiency err |
|---|---|---|---|---|---|
| 16 | 353 | 0.07 ms (err 8.2e-10) | 9.8 ms (err 5.1e-10) | **0.04 ms** | 2e-16 |
| 24 | 711 | 0.13 ms (err 8.5e-05) | 20.7 ms (err 2.9e-05) | **0.07 ms** | 2e-15 |
| 28 | 943 | 0.17 ms (err 2.2e-01) | 27.6 ms (err 2.0e-01) | **0.09 ms** | 2e-15 |
| 40 | 1656 | refused (guard) | refused (guard) | **0.16 ms** | 2e-15 |
| 100 | 3276 | refused (guard) | refused (guard) | **0.40 ms** | 9e-15 |

Same picture on real data: bioresponse trees at 32 features per path are guard-refused for the old explainers; quadrature computes SVs in 0.08 ms at 1e-16 efficiency error (order 2: 5.5 ms).

### b · Ensemble size (heloc RF depth 8, `TreeExplainer`, SV, one instance)

image

Previous backend = per-tree LinearTreeSHAP (C++, its fast path for order-1 SV) vs the new default: the whole ensemble in **one batched quadrature kernel call**, no per-tree Python objects on the hot path.

| trees | previous / point | **quadrature (batched) / point** | previous first call | quadrature first call |
|---|---|---|---|---|
| 1 | 0.04 ms | **0.02 ms** | 2 ms | 0 ms |
| 10 | 0.43 ms | **0.09 ms** | 4 ms | 2 ms |
| 50 | 2.22 ms | **0.51 ms** | 20 ms | 8 ms |
| 100 | 4.67 ms | **1.10 ms** | 38 ms | 15 ms |
| 200 | 9.51 ms | **2.20 ms** | 75 ms | 29 ms |

Per-point cost is constant in batch size (`explain_X` on the 50-tree forest, 1/10/100 points: 2.69/2.68/2.24 ms previous vs **0.76/0.68/0.55 ms** batched quadrature per point).

### c · Dataset width and interaction order

Runtime is governed by *features per decision path*, not dataset width: with a depth-12 tree, SV cost is flat from k=8 to k=81 features (0.09→0.15 ms). Any-order interactions at 12 features per path (k=32):

bench_c_features

image

| order (all interactions up to) | TreeSHAP-IQ py | Quadrature numpy | **Quadrature C++** |
|---|---|---|---|
| 2 | 136 ms | 40 ms | **0.80 ms** |
| 3 | 371 ms | 130 ms | **4.78 ms** |
| 4 | 1935 ms | 351 ms | **17.17 ms** |

### d · vs XGBoost's native TreeSHAP

image

XGBoost models explained by XGBoost itself (`pred_contribs=True`) versus the converted model through shapiq, one thread, per-instance medians. Two XGBoost baselines: **3.2.0** ships the classic TreeSHAP C++; **3.3.0** ships Wettenstein et al.'s own Quadrature-TreeSHAP (float32 SIMD, fixed 8 points, float32 output).

| config (trees×depth) | XGBoost 3.2.0 (classic) | XGBoost 3.3.0 (quadrature) | **shapiq quadrature (batched)** | shapiq linear | first call incl. conversion | max \|dev\| |
|---|---|---|---|---|---|---|
| superconductivity 100×6 | 0.35 ms | 0.21 ms | **0.30 ms** | 3.86 ms | 12 ms | 3.8e-06 |
| superconductivity 100×10 | 4.45 ms | 1.02 ms | **2.58 ms** | 7.89 ms | 34 ms | 3.8e-06 |
| superconductivity 500×8 | 5.91 ms | 2.40 ms | **4.72 ms** | 23.40 ms | 91 ms | 5.7e-06 |
| heloc 100×6 | 0.35 ms | 0.20 ms | **0.27 ms** | 2.20 ms | 10 ms | 3.0e-08 |

Against classic TreeSHAP the batched backend is at parity or faster everywhere (0.58–0.86×); against 3.3.0's float32-SIMD quadrature it is 1.5–2.2× slower — the price of exact float64 with per-depth exact point counts. This comparison also uncovered that XGBoost routes float32-cast inputs, which the converted trees now reproduce (`TreeModel.input_precision`).

### e · Sparse interaction support

The interaction output enumerates only subsets whose features co-occur on at least one decision path (all others are structurally zero and absent from the result), so wide models never materialize the dense C(F, order) enumeration. Order 1 stays a dense per-feature block. Measured on identical models, dense-output kernel vs sparse (median per explanation):

| case | interactions dense → sparse | dense | sparse | ratio |
|---|---|---|---|---|
| depth-16 tree (81 feats), SV | 81 → 81 | 1.26 ms | **0.77 ms** | 0.6× |
| depth-16 tree, SII order 2 | 3,321 → 3,320 | 3.68 ms | 7.95 ms | 2.2× |
| depth-16 tree, SII order 3 | 88,641 → 55,344 | 29.7 ms | 63.2 ms | 2.1× |
| RF 200×8 (23 feats), SV | 23 → 23 | 2.96 ms | **2.84 ms** | 1.0× |
| RF 200×8, SII order 2 | 276 → 276 | 6.13 ms | 10.1 ms | 1.6× |
| depth-12 tree (32 feats), SII order 4 | 41,448 → 18,558 | 13.9 ms | 15.8 ms | 1.1× |
| **RF 100×4, 1000 feats, SII order 3** | **166,667,500 → 89** | infeasible | **0.28 ms** | — |

The Shapley hot path is equal or faster (the dense kernel rebuilt its rank tables per call); interaction orders pay 1.6–2.2× for the sorted-table lookup (galloping cursor) — still 100×+ faster than TreeSHAP-IQ — and in exchange the last row exists at all: dense order-3 on the 1000-feature forest would need a 1.3 GB output row and a 1.7e8-entry lookup.

### Quadrature point count

Default is the exact bound ⌈d/2⌉. A fixed 8 points (paper recommendation, float32-calibrated) only pays above ~16 features per path — 13–35% faster at d=40–100 — and stayed within 2.2e-14 of the exact result everywhere measured, including degree-99 integrands at d=100. The default stays exact; `n_quadrature_points=8` is the documented speed override.

### Takeaways

- The C++ quadrature kernel is the fastest shapiq explainer at every measured setting, including ~2× vs the LinearTreeSHAP C++ kernel for plain SVs (half-size Gauss–Legendre grid, no Offset/Norm bookkeeping) — hence it is now the default path-dependent algorithm of `TreeExplainer`.
- Ensembles run in one batched kernel call: the former ~20 µs-per-tree Python dispatch is gone (200-tree forest: 2.8 ms vs 9.9 ms per explanation).
- The numpy reference implementation is ~100× slower than the C extension — it lives in the test suite as the differential oracle; the shipped explainer requires the C extension.
- The scaling variable everywhere is leaves × features-per-path; depth and dataset width only matter through those.
- On these natural datasets the fitted trees stay below the ~30 features-per-path limit of the old explainers (max observed: 27), so the numbers above are pure speed gains before the correctness advantages even come into play.

ron-wettenstein:
Read the PR Summary and the benchmarking comment (the 2 comments in this overview). This sounds good!

mmschlk:
> Read the PR Summary and the benchmarking comment (the 2 comments in this overview). This sounds good!

Really exciting to have QuadratureSHAP there as well @ron-wettenstein. Very nice work on making this numerically stable. This saves our algorithms! Note, we not fully follow your paper but mostly. 😇

src/shapiq/tree/quadrature/cext/cext.cc
```diff
@@ -0,0 +1,261 @@
...
+                }
+                if (binom > cap / factor)
+                {
+                    binom = cap;
+                    break;
+                }
+                binom = binom * factor / k;
+            }
+            expected = (binom >= cap || expected >= cap - binom) ? cap : expected + binom;
+        }
```
Advueu963: Why are we calculating the expected output. Why not trust the output array we give? Why not a sparse implementation using Hashmaps, that would then scale to any order interactions quite easily. This seems like it could blow up for order 3/4 in high depth trees.... Not sure.

src/shapiq/tree/quadrature/computer.py
```diff
@@ -0,0 +1,494 @@
...
+        self._n_features_in_tree: int = len(self._relevant_features)
+        ensemble_id = {orig: pos for pos, orig in enumerate(self._relevant_features)}
+        self._trivial_computation = self._n_features_in_tree == 0
+
+        # output lookup over the original feature ids, matching TreeSHAPIQ's packaging
+        self._interactions_lookup_relevant: dict[tuple, int] = generate_interaction_lookup(
+            self._relevant_features,
+            self._min_order,
+            self._max_order,
+        )
```
Advueu963: This could potentially be a slow function, for very deep trees and high interactin order?

src/shapiq/tree/quadrature/computer.py
```diff
@@ -0,0 +1,494 @@
...
+                    1.0 / edge_tree.p_e_values,
+                    1e-300,
+                )
+
+            def rebase(indices: np.ndarray, offset: int = node_offset) -> np.ndarray:
+                rebased = np.asarray(indices, dtype=np.int64).copy()
+                rebased[rebased >= 0] += offset
+                return rebased
+
+            roots.append(node_offset)
```
Advueu963, line 267: What is the meaning of this?

src/shapiq/tree/quadrature/computer.py
```diff
@@ -0,0 +1,494 @@
...
+                    quadrature_tree_shap,  # ty: ignore[unresolved-import]  # noqa: F401
+                )
+
+                self._cpp_available = True
+            except ImportError:
+                if implementation == "cpp":
+                    msg = "The QuadratureTreeSHAP C extension is not available in this build."
+                    raise ImportError(msg) from None
+        self._implementation = implementation
+        self._kernel_args: tuple | None = None
```
Advueu963: Why should that even be the case? Would this not then be a packaging issue on the shapiq side.

src/shapiq/tree/quadrature/computer.py
```diff
@@ -0,0 +1,494 @@
...
+        x_relevant = x_full[self._relevant_features]
+        n_players = max(x_full.shape[0], self._n_features_in_tree)
+
+        if self._trivial_computation:
+            interactions = np.zeros(0, dtype=float)
+        elif self._cpp_available and self._implementation in ("auto", "cpp"):
+            interactions = self._explain_cpp(x_relevant)
+        else:
+            interactions = self._explain_numpy(x_relevant)
+
```
Advueu963, line 368: I think really we should route to the C++ Implementation and leave the numpy implementation just as dead code or test code in the test suite.

src/shapiq/tree/precision.py
Advueu963, line 1: But would we not like to route to QuadratureSHAP in the near feature anyway. So why should we keep this?

src/shapiq/tree/treeshapiq.py
```diff
@@ -99,14 +99,26 @@ def __init__(
...
+        # validate and parse model; ensembles are handled by one child explainer per tree
+        validated_model = validate_tree_model(model, class_label=class_index)
+        self._tree_explainers: list[TreeSHAPIQ] = []
+        if len(validated_model) > 1:
+            self._tree_explainers = [
+                TreeSHAPIQ(
+                    tree, max_order=max_order, min_order=min_order, index=index, verbose=verbose
+                )
+                for tree in validated_model
+            ]
```
Advueu963, line 114: Would have thought we directly write the logic in the function, and then have just an outer loop on the logic function. But it seems this also works, just does not look very nice.

Advueu963 (approved):
Nice. Like the changes. This is an very nice rework and update of the tree computations. I think we finally should not be scared anymore at pointing people to the shapiq.tree module. Amazing Work!

mmschlk:
Let's gooo!.

PR #589, opened 2026-08-25
Remove libomp/DYLD workaround for treelite >= 4.7.1

mmschlk:
treelite 4.7.1 fixes the macOS rpath for libomp.dylib (dmlc/treelite#681), so the wheel finds Homebrew's libomp on its own:

- drop the custom RuntimeError in TreeExplainer._run_woodelf that explained the DYLD_FALLBACK_LIBRARY_PATH workaround; a treelite that cannot load libomp now surfaces its own OSError
- drop the macOS-specific DYLD_FALLBACK_LIBRARY_PATH test step in CI
- constrain treelite>=4.7.1 via [tool.uv] constraint-dependencies (not a dependency; treelite only enters through woodelf-explainer) until woodelf pins it upstream

## Public API Changes

-   [x] No Public API changes

Touches only unreleased code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,16 +151,8 @@ jobs:
                 run: brew install libomp
             -   name: Install test dependencies
                 run: uv sync --all-extras
-            -   if: matrix.os != 'macos-latest'
-                name: Run unit tests
+            -   name: Run unit tests
                 run: uv run pytest "tests/shapiq"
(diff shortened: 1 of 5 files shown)


mmschlk:
This was a very quick fix by the amazing people at treelite. So this just removes the now-obsolete error hsndling around woodelf. Right in time before release. ☺️

I am just gonna merge this in.

PR #590, opened 2026-08-26
Improving InterventionalTreeSHAPIQ and k-SII aggregation

Advueu963:
## Motivation and Context

  This PR substantially improves the `InterventionalTreeSHAPIQ` algorithm by moving the
  hot-path computations for non-boolean trees to C++.
This is achieved by a improved pre-processing via DFS that shares one tree walk across all reference samples, 
Additionally the baseline value prediction was moved to C++.
Those changes yield a ~263× end-to-end speedup.

Additionally, it vectorizes the k-SII aggregation in two stages. First, all subsets of the
  base interactions are built as a dense padded matrix and multiplied by their Bernoulli
  numbers. This leaves the matrix with 
as many repeatitions as there were supersets in the base interactions. 
Those are then again summed per subset, again vectorized (integer-encoded
  `bincount`), to obtain the final k-SII index (~12–27× faster.

  ## Public API Changes

  -   [x] No Public API changes

  ## How Has This Been Tested?

  All existing tests pass locally. 

  ## Checklist


mmschlk (commented):
I am a bit scared that the aggreagate function is now quite hard to read in the vectorized version what do you think? One thing to have is to have the old function be the test code to verify the vectorized version as it’s quite nice to read, or what is your opinion on this?

Also are these dead code:

obtain_E_R_values_point
in src/shapiq/tree/interventional/computer.py

InterventionalTreeSHAPIQ._preprocess_tree
in src/shapiq/tree/interventional/computer.py

preprocess_trees_point
in src/shapiq/tree/interventional/cext/cext.cc

mmschlk:
Checked the speed-up in my machine and even got 500x improvements lol. These improvements are just so nice. Maybe we should just start writing everything in cpp lol.

mmschlk:
> I am a bit scared that the aggreagate function is now quite hard to read in the vectorized version what do you think? One thing to have is to have the old function be the test code to verify the vectorized version as it’s quite nice to read, or what is your opponion on this?

The normal agg is the fallback for the not vectorized version that has a max scores parameter.

mmschlk (approved):
Nice improvements!

PR #591, opened 2026-08-27
Raise interventional Woodelf routing cut-off to n * m >= 100_000

mmschlk:
The n * m >= 100 cut-off predates the interventional C++ cohort-kernel
speedup, which made the shapiq implementation ~2 orders of magnitude
faster. Re-benchmarking both backends end-to-end (construction +
explain_X; sklearn random forests and XGBoost ensembles with 50-300
trees, depth 3-8, 10-50 features; SV and order-2 k-SII) puts the
break-even between n * m ~= 1e4 (small ensembles) and ~= 1e6 (large
forests). The penalty is asymmetric — below the break-even Woodelf's
fixed per-tree preprocessing makes it up to hundreds of times slower,
above it shapiq trails by a low single-digit factor near the boundary
(Woodelf still wins ~5x at n * m = 5e6) — so 1e5 balances the
worst-case absolute losses.

The cut-off now lives in the _WOODELF_INTERVENTIONAL_CUTOFF module
constant; tests exercising the auto-route monkeypatch it back to 100 to
stay cheap.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01J7VMjtBDq22vcp4Fd3DAus


Advueu963 (approved):
Nice!

PR #597, opened 2026-09-01
Feature/quadra product kernel

ron-wettenstein:
## Motivation and Context

Improve the ProductKernelExplainer.
1. Speed up the approach by more than an order of magnitude. I implemented the algorithm from the recent paper: https://arxiv.org/pdf/2605.05870. The paper reduce the approach complexity from O(n*d^3) to O(n*d^2) where d is the number of features and n is the number of reference points (the support vectors of an SVM, the training set of a Gaussian process). The new approach implementation is fully vectorized using numpy.
2. I extended the approach in the paper to more indices. The ProductKernelExplainer now support not just SV but also BV and interaction values including SII, k-SII, BII and Mobius.

See the full details, including the math behind the approach and running time results in the attached PDF below:
[product_kernel_computer.pdf](https://github.com/user-attachments/files/31700689/product_kernel_computer.pdf)

---

## Public API Changes

-   [ ] No Public API changes

ProductKernelExplainer now support more index options (BV, SII, k-SII, BII, Mobius). It also include a n_quadrature_points parameter that allow the user to use less points then the theoretically exact approach, get a very good approximation and better running times. 

The API of ProductKernelComputer was changed. It's not part of the Public API though, isn't it?

---

## How Has This Been Tested?
Added tests comparing this approach values to exact computation. Also I manually compared the values of the new approach to the ones of the old approach.

---

## Checklist

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 `ProductKernelComputer` now computes Shapley values by Gauss-Legendre quadrature of the product-game integral (Mohammadi et al., 2026) instead of the elementary-symmetric-polynomial recursion of PKeX-Shapley (Mohammadi et al., 2025).
 This reduces the cost per explained instance from `O(n d^3)` to `O(n d^2)`, where `n` is the number of reference points (the support vectors of an SVM, the training set of a Gaussian process) and `d` the number of features.
-In practice we measure speedups of 17x, 26x, 50x and 91x for models with `d = 20, 40, 80, 160` respectively.
+In practice we measure speedups of 27x, 37x, 53x and 95x on the Shapley computation for models with `d = 20, 40, 80, 160` respectively.
 
 The new `n_quadrature_points` argument on `ProductKernelExplainer` and `ProductKernelComputer` trades exactness for speed with a geometrically decaying error; it defaults to the exact bound.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## v1.7.1 (TBD)
+
+### QuadraSHAP for `ProductKernelExplainer`
+
+`ProductKernelComputer` now computes Shapley values by Gauss-Legendre quadrature of the product-game integral (Mohammadi et al., 2026) instead of the elementary-symmetric-polynomial recursion of PKeX-Shapley (Mohammadi et al., 2025).
+This reduces the cost per explained instance from `O(n d^3)` to `O(n d^2)`, where `n` is the number of reference points (the support vectors of an SVM, the training set of a Gaussian process) and `d` the number of features.
+In practice we measure speedups of 17x, 26x, 50x and 91x for models with `d = 20, 40, 80, 160` respectively.
+
+The new `n_quadrature_points` argument on `ProductKernelExplainer` and `ProductKernelComputer` trades exactness for speed with a geometrically decaying error; it defaults to the exact bound.
+
+**Public API change:** `ProductKernelComputer.compute_kernel_vectors`, `.compute_shapley_value`, `.compute_elementary_symmetric_polynomials`, and `.precompute_weights` are removed, replaced by `.compute_kernel_matrix(x)` and `.compute_shapley_values(x)`, which return the `(n, d)` kernel factors and all `d` Shapley values at once.
+`ProductKernelExplainer` itself is unchanged.
+
 ## v1.7.0 (2026-08-27)
 
 ### New and Improved Tree support.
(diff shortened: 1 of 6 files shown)


ron-wettenstein:
Few design choices I made:
1. The ProductKernelComputer class has the functions compute_values that computes SV and BV and compute_interaction_values that computes interaction values. The function compute_interaction_values can also compute single order values. I keep both functions as  compute_values is simpler and is easier to understand. The compute_interaction_values function is easy to read once compute_values is understood. Also compute_values is 30% faster on first order values compared to using compute_interaction_values. So I decided to use it.
2. We might want to add "Mobius" to ExplainerIndices as it will soon be supported here and in TreeExplainer. I avoided this for now. 
3. I raise warning if both than 10,000 interactions are about to be created per instance.
4. The fully vectorized algorithm is more RAM hungry. I use batching to limit the RAM usage. It can reach up to 100MB and not more. If you want to limit the RAM further we can decrease the batching constants.

ron-wettenstein:
## Results

### Setup

Synthetic product-kernel models, with the number of reference points held fixed at `n = 120`
and the feature count `d` swept, so the measurement isolates the `d` dependence:

- `X_train ~ N(0, 1)` of shape `(n, d)`, `alpha ~ N(0, 1)` of length `n`, `gamma = 0.3`
- explained point `x ~ N(0, 1)`, all drawn from `np.random.default_rng(0)`
- quadrature in the **exact** regime, `m_q = ceil(d / 2)` nodes (no approximation)

Both paths are timed **end to end from `x`**, including their own kernel construction. Each
configuration gets a warm-up call, then adaptive repetitions targeting ~1.5 s, reported as the
minimum of 3 such batches.

Environment: AMD Ryzen 7 5700U (8C/16T), Windows 11, Python 3.12.7, NumPy 2.3.3.

### Quadrature versus the elementary-symmetric-polynomial recursion

| `d` | `m_q` | recursion | quadrature | speedup | max rel. diff. |
|----:|------:|----------:|-----------:|--------:|---------------:|
|  20 |    10 |   14.37 ms |    0.40 ms |   36.0x |       5.6e-16 |
|  40 |    20 |   76.47 ms |    1.73 ms |   44.3x |       9.6e-15 |
|  80 |    40 |  440.03 ms |   10.36 ms |   42.5x |       4.1e-14 |
| 160 |    80 | 3156.08 ms |   33.34 ms |   94.7x |       1.2e-15 |
| 320 |   160 | 23856.48 ms |  134.14 ms |  177.8x |       2.9e-13 |

The last column is the largest relative difference between the two implementations' output
vectors. At every `d` they agree to floating-point noise, and on the models small enough to
enumerate both also match `ExactComputer` over all `2^d` coalitions.

### Where the speedup comes from

| `d` | recursion: kernel | recursion: Shapley | quadrature: kernel | quadrature: Shapley | Shapley-only speedup |
|----:|------------------:|-------------------:|-------------------:|--------------------:|---------------------:|
|  20 |    4.42 ms |    10.03 ms |   0.017 ms |    0.38 ms |  26.8x |
|  40 |    8.96 ms |    67.85 ms |   0.028 ms |    1.83 ms |  37.1x |
|  80 |   19.91 ms |   447.70 ms |   0.050 ms |    8.46 ms |  52.9x |
| 160 |   36.52 ms |  3155.20 ms |   0.101 ms |   33.09 ms |  95.3x |

Two independent effects:

1. **Kernel construction**, 260-360x faster. The recursion called
   `sklearn.metrics.pairwise.rbf_kernel` once per feature on a single-column slice, paying `d`
   rounds of per-call overhead; the replacement is one `exp(-gamma * (X_train - x)**2)` on the
   whole `(n, d)` array. A large constant, not an asymptotic win, so it dominates the total
   only at small `d` (31% of the recursion's runtime at `d = 20`, ~1% at `d = 160`).
2. **The Shapley computation**, 27-95x faster and *growing with `d`* -- this is the algorithmic
   change, `O(n d^3)` to `O(n d^2)`.

### Scaling

Least-squares fits of `t ~ d^p` over the swept range:

| quantity | fitted | theory |
|---|---:|---:|
| recursion, Shapley only | `d^2.76` | `d^3` |
| quadrature, Shapley only | `d^2.16` | `d^2` (since `m_q = ceil(d/2)`) |
| recursion, kernel construction | `d^1.03` | `d^1` |
| quadrature, kernel construction | `d^0.85` | `d^1` |

Both fitted exponents sit slightly below theory because the small-`d` end is dominated by
constant overheads. The successive-doubling ratios converge on the predicted values from below:
the recursion gives `d^2.82` from `d = 80` to `160` and `d^2.92` from `160` to `320`; quadrature
gives `d^1.97` and then `d^2.01` over the same steps.

### Cost per index

All five indices, timed in a single batch so the numbers are directly comparable. `SII`, `BII`
and `Moebius` use `min_order = 1`, `max_order = 2`.

| `d` | `SV` | `BV` | `SII` (1-2) | `BII` (1-2) | `Moebius` (1-2) |
|----:|-----:|-----:|------------:|------------:|----------------:|
|  20 |  0.37 ms | 0.08 ms |    2.99 ms |  1.08 ms |  0.97 ms |
|  40 |  1.33 ms | 0.13 ms |   30.94 ms |  4.14 ms |  3.99 ms |
|  80 |  8.24 ms | 0.22 ms |  334.58 ms | 16.28 ms | 15.45 ms |
| 160 | 32.86 ms | 0.83 ms | 2835.80 ms | 80.77 ms | 79.26 ms |

PR #598, opened 2026-09-11
fix(tests): pass background_reg_dataset to the sequential model fixtures

Anai-Guo:
### What

The three Keras fixtures `sequential_model_1_class`, `sequential_model_2_classes` and
`sequential_model_3_classes` call `_sequential_model()` with one argument, but that helper
has required **two** parameters since #334:

```python
def _sequential_model(output_shape_nr, background_reg_dataset) -> Model:
```

so every one of them raises

```
TypeError: _sequential_model() missing 1 required positional argument: 'background_reg_dataset'
```

### Why it matters

All three names are listed in `TABULAR_TENSORFLOW_MODEL_FIXTURES`, which
`tests/shapiq/conftest.py` folds into `ALL_MODEL_FIXTURES`, and
`test_tensorflow_get_predict_function_and_model_type` resolves them through
`request.getfixturevalue(model_name)`. Whenever that test actually runs, all three
parameterisations error out.

Note that the argument binding fails *before* the body executes, so the
`pytest.importorskip("keras")` guard inside `_sequential_model` never gets a chance to run —
the fixtures raise `TypeError` rather than skipping cleanly, even on machines without Keras.
The test is gated behind `@pytest.mark.skipif(... "tensorflow" not in sys.modules)` and
`@pytest.mark.external_libraries`, which is why this has gone unnoticed.

### Origin

#334 converted the model fixtures to take their dataset from a fixture parameter, e.g.

```diff
-def xgb_reg_model():
+def xgb_reg_model(background_reg_dataset) -> Model:
-def dt_reg_model() -> DecisionTreeRegressor:
+def dt_reg_model(background_reg_dataset) -> DecisionTreeRegressor:
```

and changed `_sequential_model` to require `background_reg_dataset`. For these three
fixtures only the `-> Model` return annotation was added and the parameter was missed.
Every other model fixture in the file already declares it.

### Fix

Declare `background_reg_dataset` on the three fixtures and forward it, matching the
sibling fixtures.

### Verification

Reduced the fixtures to a standalone file preserving the exact structure (three no-arg
fixtures → one-arg call → two-param helper with `importorskip`, driven through
`request.getfixturevalue`):

```
BEFORE:  3 failed  — TypeError: _sequential_model() missing 1 required positional
                     argument: 'background_reg_dataset'
AFTER:   3 skipped — could not import 'keras': No module named 'keras'
```

i.e. before the change the fixtures error regardless of the environment; after it they
skip correctly when Keras is absent and build the model when it is present.

`ruff check` and `ruff format --diff` are clean on the changed file with the repo's
`pyproject.toml` (pinned `ruff==0.9.10` from `.pre-commit-config.yaml`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Advueu963:
Hey Anai-Guo,

Thanks for the PR!
I was wondering if we are still using these fixtures anywhere?
Because when I did a quick search, I did not find them used at all, which would also hint at why the tests worked the entire time until now.
So it might be best to either (A) add tests for those (not sure we need them though) or (B) remove the fixtures.

Thanks for helping to make shapiq better!

Best,
Santo

